PR #1, opened 2013-12-19
Added support for fetch_all_like queries to BaseAdaptor::AUTOLOAD

swingingsimian:
Added support for fetch_all_like queries to BaseAdaptor::AUTOLOAD. Throws if used with fetch_like as opposed to fetch_all_like, and similarly throws if there is more than 1 filter component.


ens-lg4:
Dear Nathan,

Sorry for keeping you waiting for so long.

Unfortunately I cannot accept this pull request both for a technical reason (it cannot be merged) and because it is a change in the low-level API that may encourage wrong style of programming from the potential users of this code.

I believe that *if* global variables are used by the pipeline, their exact names should be known to the code.

PR #7, opened 2014-10-23
Feature/runnabledb tests

kiwiroy:
A useful testsuite?


muffato:
Hi Roy,

As you could see, we have finally merged this branch. This has revealed to be a very interesting task. We have done the following to your branch:
 - most of init_pipeline.pl, standaloneJob.pl and runWorker.pl have moved to some Perl modules that can be called from both the scripts themselves and the test interface.
 - new Bio::EnsEMBL::Hive::Utils::Tests that exports wrappers for those 3 scripts
 - no need to have t/lib/Hive/, which limits the overhead due to the test suite (and we're sure that we're testing the right bits)
 - we test the three long-multiplication pipelines by running all their jobs and checking that the output is correct

Thanks a lot for your contribution
Matthieu

PR #10, opened 2015-08-26
Update FastaFactory.pm

mn1:
This works for me.


muffato:
Hi Michael,

What do you want to achieve with this change ? We can see a couple of problems:
 - you've made chunk_name equal to chunk_number. Why that ?
 - output_prefix must now be defined to a proper directory. It can't be a filename prefix any more
We can see, though, that mkpath should indeed be called on the whole directory (incl. any prefix).
Maybe there should be another option to control the output directory (on top of output_prefix) ?

mn1:
Based on our discussion I'll close the pull request. I'll include an output_dir parameter and create another pull request once it is ready.

mn1:
What about changing the name of the parameter output_prefix to file_prefix?

Either way is ok with me, but I think file_prefix is more intuitive to use.

PR #11, opened 2015-08-26
Update FastaFactory.pm

mn1:
Added output_dir parameter for specifying directories, it defaults to the current directory. ('.')

chunk_name in the output now has the full path to the fasta file.

output_prefix is a prefix to the file name.

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
+++ b/modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
@@ -80,6 +80,7 @@ sub param_defaults {
         'output_prefix'     => 'my_chunk_',
         'output_suffix'     => '.fasta',
         'hash_directories'  => 0,
+        'output_dir'        => '.',
     };
 }
 
@@ -130,26 +131,35 @@ sub write_output {
 
     my $input_seqio         = $self->param('input_seqio');
     my $max_chunk_length    = $self->param('max_chunk_length');
+    my $output_dir          = $self->param('output_dir');
     my $output_prefix       = $self->param('output_prefix');
     my $output_suffix       = $self->param('output_suffix');
 
     my $chunk_number = 1;   # counts the chunks
     my $chunk_length = 0;   # total length of the current chunk
     my $chunk_size   = 0;   # number of sequences in the current chunk
-    my $chunk_name   = $output_prefix.$chunk_number.$output_suffix;
+    
+    my $current_output_directory;    
     if ($self->param('hash_directories')) {
         my $dir_tree = dir_revhash($chunk_number);
-        if ($dir_tree ne '') {
-            mkpath($dir_tree);
-            $chunk_name = $dir_tree.'/'.$chunk_name;
+        
+        if ($dir_tree eq '') {
+	  $current_output_directory = $output_dir;
+        } else {
+	  $current_output_directory = $output_dir . '/' . $dir_tree;
         }
+    } else {      
+      $current_output_directory = $output_dir;
     }
-    my $chunk_seqio  = Bio::SeqIO->new(-file => '>'.$chunk_name, -format => 'fasta');
+    mkpath($current_output_directory);
+    my $chunk_name   = $current_output_directory . '/' . $output_prefix . $chunk_number . $output_suffix;
+    my $chunk_seqio  = Bio::SeqIO->new(-file => '>' . $chunk_name, -format => 'fasta');
     
-    while (my $seq_object = $input_seqio->next_seq) {
-	$chunk_seqio->write_seq( $seq_object );
+    while (my $seq_object = $input_seqio->next_seq) {	
 	
         if((my $seq_length = $seq_object->length()) + $chunk_length <= $max_chunk_length) {
+        
+	    $chunk_seqio->write_seq( $seq_object );
             $chunk_length += $seq_length;
             $chunk_size   += 1;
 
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -130,26 +131,35 @@ sub write_output {
...
+    my $chunk_name   = $current_output_directory . '/' . $output_prefix . $chunk_number . $output_suffix;
+    my $chunk_seqio  = Bio::SeqIO->new(-file => '>' . $chunk_name, -format => 'fasta');
     
-    while (my $seq_object = $input_seqio->next_seq) {
-	$chunk_seqio->write_seq( $seq_object );
+    while (my $seq_object = $input_seqio->next_seq) {	
 	
         if((my $seq_length = $seq_object->length()) + $chunk_length <= $max_chunk_length) {
+        
+	    $chunk_seqio->write_seq( $seq_object );
```
muffato: You can't see it (tests are only available since 2.3), but this would break the test-suite. The problem is that you moved `write_seq()` into the `if`, which causes the runnable to output empty files (if the sequences are bigger than the chunk size). Note that the runnable does *not* split the actual sequences, it only groups the sequences in a different manner

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -167,15 +177,22 @@ sub write_output {
...
+                } else {
+                  $current_output_directory = $output_dir . '/' . $dir_tree;
                 }
+                mkpath($current_output_directory);
+            } else {            
+	      $current_output_directory = $output_dir;
             }
-            $chunk_seqio    = Bio::SeqIO->new(-file => '>'.$chunk_name, -format => 'fasta');
+            $chunk_name   = $current_output_directory . '/' . $output_prefix . $chunk_number . $output_suffix;
+            $chunk_seqio = Bio::SeqIO->new(-file => '>' . $chunk_name, -format => 'fasta');
```
muffato: This line seems identical in function to what was line 178. Can you restore the original indentation ?

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -130,26 +131,35 @@ sub write_output {
...
+        } else {
+	  $current_output_directory = $output_dir . '/' . $dir_tree;
         }
+    } else {      
+      $current_output_directory = $output_dir;
     }
-    my $chunk_seqio  = Bio::SeqIO->new(-file => '>'.$chunk_name, -format => 'fasta');
+    mkpath($current_output_directory);
+    my $chunk_name   = $current_output_directory . '/' . $output_prefix . $chunk_number . $output_suffix;
+    my $chunk_seqio  = Bio::SeqIO->new(-file => '>' . $chunk_name, -format => 'fasta');
```
muffato: This line seems identical in function to what was line 147. Can you restore the original indentation ?

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -130,26 +131,35 @@ sub write_output {
...
     
-    while (my $seq_object = $input_seqio->next_seq) {
-	$chunk_seqio->write_seq( $seq_object );
+    while (my $seq_object = $input_seqio->next_seq) {
```
muffato: This line seems identical in function to what was line 149. Can you restore the original indentation / spacing ?

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -130,26 +131,35 @@ sub write_output {
...
-            mkpath($dir_tree);
-            $chunk_name = $dir_tree.'/'.$chunk_name;
+        
+        if ($dir_tree eq '') {
+	  $current_output_directory = $output_dir;
+        } else {
+	  $current_output_directory = $output_dir . '/' . $dir_tree;
         }
+    } else {      
+      $current_output_directory = $output_dir;
```
muffato: The block can probably look cleaner if `$current_output_directory` is initialized to `$output_dir` and modified when needed

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -167,15 +177,22 @@ sub write_output {
...
-                    mkpath($dir_tree);
-                    $chunk_name = $dir_tree.'/'.$chunk_name;
+                if ($dir_tree eq '') {                
+		  $current_output_directory = $output_dir;
+                } else {
+                  $current_output_directory = $output_dir . '/' . $dir_tree;
                 }
+                mkpath($current_output_directory);
+            } else {            
+	      $current_output_directory = $output_dir;
```
muffato: The block can probably look cleaner if `$current_output_directory` is initialized to `$output_dir` and modified when needed

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
mn1: The reason why I moved the write_seq statement was that in the original version $chunk_length and $chunk_size were wrong, because the last sequence written to the file, but $chunk_length and $chunk_size weren't updated to account for that.

Point taken about the empty files. Even worse, my version was omitting the last sequence. I'm fixing this.

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
mn1: Agreed. Done.

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
muffato: oh sh**, you're right, chunk_size and chunk_length are 0 in some occasions. The unit-test was only checking that when the chunk_size is greater than the combined length of all the sequences. I'll fix that now (it has to be fixed on 1.9 I guess and merged into all the following branches)

muffato: Now fixed on all the branches (2c545d1)

PR #12, opened 2015-12-18
Update init_pipeline.pl

tweep:
Using Getopt::ArgvFile allows to pass long, complex commandline options via a file to init_pipeline.pl, so you can overwrite various options in your PipeConfig module without having super-long commandlines. We needed this as we run various pipelines with ever-changing options, and we do not want to change the installation of our pipelines all the time. 
Getopt::ArgvFile plays well with Getopt::Long and no additional changes are needed - well, except installing Getopt::ArgvFile ... 

USAGE: perl init_pipeline.pl Bio::EnsEMBL::Hive::PipeConfig:;Whaterver   @myconfigfile.conf 

The format of the configuration file you pass in is a key/value file :
--option value 
--option2 value2

### DIFF
--- a/scripts/init_pipeline.pl
+++ b/scripts/init_pipeline.pl
@@ -13,7 +13,9 @@ BEGIN {
     unshift @INC, $ENV{'EHIVE_ROOT_DIR'}.'/modules';
 }
 
-
+use Getopt::ArgvFile; # Enable passing of long commandline options by file. 
+                      # Example: perl init_pipeline.pl B::E::H::P::Whaterver_conf  @my_file.conf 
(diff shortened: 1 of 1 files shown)


dbolser:
Looks good, but couldn't you just create a custom _conf.pl that inherits
the pipe conf and put all your custom settings there?

muffato:
Hi @tweep ,

I have accepted the request but with a different implementation.
Getopt::Argvfile is kept optional (I don't want to force the extra dependency), but the code dies if you use the @ syntax and Getopt::Argvfile is not installed

tweep:
Thank you Matthieu !

PR #14, opened 2016-01-27
Update LSF.pm to parse year- and year-less responses of LSF bacct cmd

tweep:
Our installation of LSF returns the year in the bacct response - which is not parsed by the regex in parse_report_source_line(). The fix will parse year- and year-less responses of bacct. This is a quickfix as the _yearless_2_datetime() method is  still invoked - best would be to not invoke it if the year is returned; however I think it increases the complexity of the code too much.
Regex now parses both lines: 

```Sat Jan 23 22:48:28 2016: Completed <exit>; TERM_RUNLIMIT: job killed after reaching LSF run time limit.```

```Sat Jan 23 22:48:28: Completed <exit>; TERM_RUNLIMIT: job killed after reaching LSF run time limit.```

screen shot 2016-01-26 at 4 06 05 pm

We're using LSF version 9.1.2.0 - would be interesting if the different response formats are due to different LSF versions. Maybe it's time to create Bio::EnsEMBL::Hive::Meadow::LSF::SUBVERSION ?

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Meadow/LSF.pm
+++ b/modules/Bio/EnsEMBL/Hive/Meadow/LSF.pm
@@ -224,7 +224,8 @@ sub parse_report_source_line {
             my (@keys, @values);
             my $line_has_key_values = 0;
             foreach (@lines) {
-                if( /^(\w+\s+\w+\s+\d+\s+\d+:\d+:\d+):\s+Completed\s<(\w+)>(?:\.|;\s+(\w+))/ ) {
+                // Parse yearless + year bacct response
+                if( /^(\w+\s+\w+\s+\d+\s+\d+:\d+:\d+)(?:\s*\d*):\s+Completed\s<(\w+)>(?:\.|;\s+(\w+))/ ) {
                     $when_died      = _yearless_2_datetime($1);
                     $cause_of_death = $3 && $status_2_cod{$3};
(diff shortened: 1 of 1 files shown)


ens-lg4:
Hi Jan,

Thanks for reporting this issue. Very cool illustration of the problem!

We definitely want to support LSF 9.1.2 although we currently run LSF 9.1.1.1 . But I don't think we are ready to introduce separate Meadow modules for each version as you suggested. I think a generalized approach similar to what you proposed is enough.

Please fix a few issues with the current pull request and we should be able to accept it:

(1) C++ style comments that broke the Travis unit test (you probably got a broken build notification already) - just use # instead of //

(2) A year is not a random number of digits - it should be exactly four after some space.

(3) You are correct - we don't need to call _yearless_2_datetime() function if the year has been captured. I think the best solution is to attempt to capture the 4-digit year (without spaces), and if it's not defined, then call _yearless_2_datetime() , for example:
    $when_died = $4 // _yearless_2_datetime($1);

(4) Ideally (since it is a bug fix) the patch should go to version/1.9 branch, from where we will cascade-merge it all the way up. But if it's too tricky - leave it on the master and I will rebase it.

Cheers,

Leo.

muffato:
This has been fixed in separate commits by @ens-lg4

PR #20, opened 2016-05-12
Fix for DS slow registry

muffato:
This fixes @danstaines's issue:
[Feb 11th at 09:49]
I've got a problem with hive where I supply a registry, and the registry takes a couple of minutes to load (this is normal with 40k bacteria unfortunately).

The registry takes a lot of time to load, so the worker is running from the LSF point-of-view but is not registered in the database yet (because the DBAdaptor object is not ready yet). Because of the discrepancy, beekeeper thinks that nothing is running and keep on submitting new workers (beyond the analysis capacity)

My fix is to basically count those workers as _pending_. Because the Valley/Meadow had a method to get the running workers and another one to get the pending workers, I had to unify them into a single method that takes the list of the workers that have registered and return the correct counts. This super-hash is then carried around in various places so that all the decisions are consistent to each other. Overall the Valley now does more things and the Meadows are a bit shorter

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Meadow.pm
+++ b/modules/Bio/EnsEMBL/Hive/Meadow.pm
@@ -203,48 +203,16 @@ sub get_current_worker_process_id {
 }
 
 
-=head2 count_pending_workers_by_rc_name
-
-    Title   :  count_pending_workers_by_rc_name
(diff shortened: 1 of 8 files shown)


muffato:
Hum, I haven't updated the test-suite, so it's going to fail ... I'll fix that but I'd still like another pair of eyes on this diff

PR #21, opened 2017-01-10
Use File::Glob to get the list of filenames

Zhicheng-Liu:
Previously backtick ls command was used to fetch the list of sql patches filenames.
By using backtick it is required to fork a new process. This would potentially be
a problem for a running process that uses large amount of memory. If there is not
enough spare memory left in the system, the fork operation could be aborted because
of "cannot allocate memory" error.

Fixes ENSEMBL-4823.

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/DBSQL/SqlSchemaAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Hive/DBSQL/SqlSchemaAdaptor.pm
@@ -38,12 +38,14 @@ package Bio::EnsEMBL::Hive::DBSQL::SqlSchemaAdaptor;
 use strict;
 use warnings;
 
+use File::Glob qw(bsd_glob);
+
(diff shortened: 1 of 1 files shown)


muffato:
This patch works on my laptop but not on the Sanger and EBI farms. On the latter, I need to remove the single quotes from the `glob` pattern, and then it works on all.
@Zhicheng-Liu : can you try this and check it works for you ?
This test especially: `prove t/02.api/sql_version.t`

Zhicheng-Liu:
@muffato : thank you for pointing out the single quote issue. I have removed single quotes inside the double quotes. I checked it on ebi-001, gunpowder and ves-hx-60 and it worked on all of them.

muffato:
lsf.t is expected to fail here. I could cherry-pick the bugfix I've put on master, but I'm not sure it's worth it. Otherwise, it looks good to me

PR #22, opened 2018-01-31
Master

mateuspatricio:
New changes in FastaFactory, allowing it to filter out unwanted sequences and to split a FASTA file by the amount of sequences.

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
+++ b/modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
@@ -29,6 +29,8 @@
 
         param('max_chunk_length');  # Maximum total length of sequences in a chunk: 'max_chunk_length' => '200000'
 
+        param('max_chunk_size');    # Defines the maximum allowed number of sequences to be included in each output file.
+
         param('seq_filter');        # Can be used to exclude sequences from output files. e.g. '^TF' would exclude all sequences starting with TF.
 
         param('output_prefix');     # A common prefix for output files: 'output_prefix' => 'my_special_chunk_'
@@ -89,6 +91,7 @@ sub param_defaults {
 
     return {
         'max_chunk_length'  => 100000,
+        'max_chunk_size'    => 0,
         'output_prefix'     => 'my_chunk_',
         'output_suffix'     => '.#input_format#',
         'seq_filter'        => undef,
@@ -152,6 +155,7 @@ sub write_output {
 
     my $input_seqio         = $self->param('input_seqio');
     my $max_chunk_length    = $self->param('max_chunk_length');
+    my $max_chunk_size      = $self->param('max_chunk_size');
     my $output_prefix       = $self->param('output_prefix');
     my $output_suffix       = $self->param('output_suffix');
     my $output_dir          = $self->param('output_dir');
@@ -178,7 +182,7 @@ sub write_output {
         $chunk_length += $seq_object->length();
         $chunk_size   += 1;
 	
-        if ($chunk_length > $max_chunk_length) {
+        if (($max_chunk_length && ($chunk_length > $max_chunk_length)) or ($max_chunk_size && ($chunk_size > $max_chunk_size))) {
 
                 # dataflow the current chunk:
             $self->dataflow_output_id( {
--- a/modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
+++ b/modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
@@ -29,6 +29,8 @@
 
         param('max_chunk_length');  # Maximum total length of sequences in a chunk: 'max_chunk_length' => '200000'
 
+        param('seq_filter');        # Can be used to exclude sequences from output files. e.g. '^TF' would exclude all sequences starting with TF.
+
         param('output_prefix');     # A common prefix for output files: 'output_prefix' => 'my_special_chunk_'
 
         param('output_suffix');     # A common suffix for output files: 'output_suffix' => '.nt'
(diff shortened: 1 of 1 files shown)


mateuspatricio:
Hi @ens-bwalts,
I added a description of the differences of these two variables. In addition to an example on how the filter could be used.
Mateus.

muffato:
I think the description should be in the POD, not in the code

mateuspatricio:
Yes, good point. Moving it now.

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -28,6 +28,10 @@
         param('inputfile');         # The original Fasta file: 'inputfile' => 'my_sequences.fasta'
 
         param('max_chunk_length');  # Maximum total length of sequences in a chunk: 'max_chunk_length' => '200000'
+	
+	param('max_chunk_size');    # Defines the maximum allowed number of sequences to be included in each output file.
+	
+	param('seq_filter');        # Can be used to exclude sequences from output files. e.g. '^TF' would exclude all sequences starting with TF.
```
muffato: Should be spaces, not tabs

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -167,11 +173,13 @@ sub write_output {
     
     while (my $seq_object = $input_seqio->next_seq) {
 
+        next if ( $seq_object->id =~ /$seq_filter/ );
```
muffato: In fact: can only do that if `$seq_filter` is defined

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
muffato: Should be spaces, not tabs

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -148,6 +152,7 @@ sub write_output {
 
     my $input_seqio         = $self->param('input_seqio');
     my $max_chunk_length    = $self->param('max_chunk_length');
+    my $max_chunk_size      = $self->param('max_chunk_size');
```
muffato: Need to define a default value in `param_defaults`

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -156,6 +161,7 @@ sub write_output {
     my $chunk_length = 0;   # total length of the current chunk
     my $chunk_size   = 0;   # number of sequences in the current chunk
     my $chunk_name   = $output_prefix.$chunk_number.$output_suffix;
+    my $seq_filter   = $self->param('seq_filter');
```
muffato: Need to define a default value in `param_defaults`

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
mateuspatricio: Done, tabs came from editing the file on the browser.

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
mateuspatricio: Done.

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
mateuspatricio: Thats correct. Otherwise an empty regex would match all IDs and exclude all sequences. Fixed now.

muffato:
Still waiting for some Travis jobs but several jobs are already failing. It seems `latest_schema_patch.t` is the culprit, which may have nothing to do with Mateus' change itself. The test checks out different versions of the schema using git tags. Perhaps these are not checked out on Travis when it builds off a pull-request ?

ens-bwalts:
Looks like muffato has it right. It's grabbing ensembl-hive using a git clone --depth=500. If I do that in a shell, I only get tags through sql_schema_80_start. If I clone without the depth, then I get all the sql_schema_N_start tags.

So, why does it work normally? When it pulls from the main (non-pull-request) branch, it does the clone with --depth=500 _and_ --branch=master (or whichever). In that case, I get all the tags.

--depth can be changed in .travis.yml. See https://docs.travis-ci.com/user/customizing-the-build/#Git-Clone-Depth - we should probably protect that change with a - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; so we're not doing it with every build (https://docs.travis-ci.com/user/pull-requests/)

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -28,6 +28,10 @@
         param('inputfile');         # The original Fasta file: 'inputfile' => 'my_sequences.fasta'
 
         param('max_chunk_length');  # Maximum total length of sequences in a chunk: 'max_chunk_length' => '200000'
+
```
muffato: Need to remove the tab here

muffato: Need to remove the tab here

modules/Bio/EnsEMBL/Hive/RunnableDB/FastaFactory.pm
```diff
@@ -28,6 +28,10 @@
         param('inputfile');         # The original Fasta file: 'inputfile' => 'my_sequences.fasta'
 
         param('max_chunk_length');  # Maximum total length of sequences in a chunk: 'max_chunk_length' => '200000'
+	
+        param('max_chunk_size');    # Defines the maximum allowed number of sequences to be included in each output file.
+
```
muffato: Need to remove the tab here

muffato:
I'm running the test-suite locally to make sure the Travis failure was not due to the change in FastaFactory

PR #23, opened 2018-03-09
More options when running system commands

muffato:
I've added a few options in `Process::run_system_command` and `RunnaleDB::SystemCmd` to:
- add a time limit (mimicking LSF's _RUNLIMIT_)
- capture the standard output
- record how long the command took

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/RunnableDB/SystemCmd.pm
+++ b/modules/Bio/EnsEMBL/Hive/RunnableDB/SystemCmd.pm
@@ -141,17 +141,36 @@ sub write_output {
         }
 
         if ($stderr =~ /Exception in thread ".*" java.lang.OutOfMemoryError: Java heap space at/) {
-            my $job_ids = $self->dataflow_output_id( undef, -1 );
-            if (scalar(@$job_ids)) {
-                $self->input_job->autoflow(0);
-                $self->complete_early("Java heap space is out of memory. A job has been dataflown to the -1 branch.\n");
-            } else {
-                die $stderr;
-            }
+            $self->complete_early_if_branch_connected("Java heap space is out of memory. A job has been dataflown to the -1 branch.\n", -1);
+            die $stderr;
         }
 
         die sprintf( "'%s' resulted in an error code=%d\nstderr is: %s\n", $flat_cmd, $return_value, $stderr);
     }
+}
+
+
+######################
+## Internal methods ##
+######################
+
+=head2 complete_early_if_branch_connected
+
+  Arg[1]      : (string) message
+  Arg[2]      : (integer) branch number
+  Description : Wrapper around complete_early that first checks that the
+                branch is connected to something.
+  Returntype  : void if the branch is not connected. Otherwise doesn't return
+
+=cut
+
+sub complete_early_if_branch_connected {
+    my ($self, $message, $branch_code) = @_;
+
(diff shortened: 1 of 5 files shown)


modules/Bio/EnsEMBL/Hive/RunnableDB/SystemCmd.pm
```diff
@@ -137,23 +144,59 @@ sub write_output {
...
+
+  Arg[1]      : (integer) branch number
+  Arg[2]      : (string) message
+  Description : Wrapper around _complete_early_with_single_dataflow that first
+                checks that the branch is connected to something.
+  Returntype  : 0 if the branch is not connected. Otherwise doesn't return
+
+=cut
+
+sub _dataflow_if_branch_connected {
```
ens-bwalts: Have you considered moving _dataflow_if_branch_connected and _complete_early_with_single_dataflow up into Process.pm? They may be useful in other runnables, even user-developed ones.

muffato: Yes I thought about that too.
 - `complete_early_with_single_dataflow` is indeed a construct that we do many times in Compara. I guess it'd be useful too for others. I was slightly worried that it could be confusing to have it together with `complete_early` but I could perhaps merge `complete_early_with_single_dataflow` into `complete_early`:
```perl
sub complete_early {
    my ($self, $message, $branch_code) = @_;
    if (defined $branch_code) {
        $self->dataflow_output_id(undef, $branch_code);
        $self->input_job->autoflow(0);
    }
    $self->input_job->incomplete(0);
    die $message;
}
```
Which way do you prefer ?
- `dataflow_if_branch_connected`. About this one, I had a "philosophical" concern. Ideally, Runnables should be agnostic of the pipelines they're plugged in, and how they're wired. In reality they cannot be fully disconnected, but at least they shouldn't be _guessing_ it. I'd prefer a more declarative approach like https://github.com/Ensembl/ensembl-compara/blob/master/modules/Bio/EnsEMBL/Compara/RunnableDB/ncRNAtrees/SecStructModelTree.pm#L128 where the analysis has a parameter that tells it whether there is an "escape" branch or not. Also, my implementation of `dataflow_if_branch_connected` only checks whether there is a _rule_ attached to the branch, ignoring that there can be conditional targets and that the dataflow may not end up writing anything to the database. Here my assumption is that if the runnable has a -1 or -2 branch connected to it, then it's probably fine to flow MEMLIMIT and RUNLIMIT events to these, rather than having two extra parameters `can_detect_memlimit` and `can_detect_runlimit`. I think my main comment is that I'd like to keep `dataflow_output_id` the clear preferred way of dataflowing between runnnables.

ens-bwalts: I like combining complete_early_with_single_dataflow with complete_early

As for dataflow_if_branch_connected: the point about coupling the Runnable with the pipeline is a good one. The additional constraints that would be forced by having dataflow_if_branch_connected implemented would not be good in all cases. Also, I agree with your point that it's specialized to check for branch -1 or -2 dataflow - offering it as something to work with any dataflow branch (without checking for conditionals, or even the type of dataflow target) would be inviting problems.

muffato:
I have updated the commits based on combining `_complete_early_with_single_dataflow` with `complete_early`

PR #26, opened 2018-03-21
New unit-test for dataflow_output_id under various conditions

muffato:
This is a request for comments. I've tried to compile a list of all possible `dataflow_output_id` calls, with the expected behaviour from Hive. At the moment, the test doesn't pass on 2.4 (neither it does on master), but I've got a couple of commits ready, assuming these below are the expected outcomes.

There is a case where I really don't know what's more "natural". If the template needs `a_multiplier` and `a_multiplier` is set to 9650156169 in the input_id, but to 27 in the parameter hash (the parameter has been modified at runtime). Which value should be used in the dataflow ??

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/AnalysisJob.pm
+++ b/modules/Bio/EnsEMBL/Hive/AnalysisJob.pm
@@ -284,7 +284,17 @@ sub dataflow_output_id {
     my $input_id                = $self->input_id();
     my $hive_use_param_stack    = $self->hive_pipeline->hive_use_param_stack;
 
+    $output_ids    =  destringify($output_ids) unless ref($output_ids);     # destringify the string
     $output_ids    = [ $output_ids ] unless(ref($output_ids) eq 'ARRAY');   # force previously used single values into an arrayref
+
+    my @destringified_output_ids;
+    foreach my $output_id (@$output_ids) {
+        $output_id = destringify($output_id) unless ref($output_id);        # destringify the string
+        if ((defined $output_id) and (ref($output_id) ne 'HASH')) {         # Only undefs and hashrefs work as input_ids
+            die stringify($output_id)." is not a hashref ! Cannot dataflow";
+        }
+        push @destringified_output_ids, $output_id;
+    }
 
         # map branch names to numbers:
     my $branch_code = Bio::EnsEMBL::Hive::DBSQL::DataflowRuleAdaptor::branch_name_2_code($branch_name_or_code);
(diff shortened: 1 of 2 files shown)


ens-lg4:
I'd like to offer an opinion, if I may.
Because all substitutions happen at run time, the current state of all params should have precedence over anything that was set at job init time.  This is what allows us to build flexible factories.

muffato:
My question was: do parameters (values) that are neither in the output_id nor the input_id be considered when evaluating templates ?

Imagine a simple pipeline analysis A -> analysis B (autoflow)
At the moment, if A is given alpha=23 in its input_id, but changes it to 45 at runtime:
 - If the dataflow rule has no template, B gets alpha=23
 - If the dataflow rule has a `{'alpha' => '#alpha#}'` template, B gets alpha=45
The template forces the dataflow to use the freshest value available

ens-bwalts:
I agree with Leo. If a runnable updates a parameter value, the new value should be propagated regardless of whether or not there is a template in the dataflow rule. More specifically, the value that the seeded job sees should be the value at the instant the dataflow event happens.

ens-lg4:
I think I understand the problem better after Matthieu's explanation.  It is indeed counter-intuitive that this "refreshing" of the value happens due to presence of a template.  Yet I would still like to have the option to access the runtime values. On the other hand, if there is no explicit dataflow we would expect the original values to propagate...

It is tempting, for the sake of clarity and flexibility, to extend the language once again and make both #alpha._input# and #alpha._runtime# available to the pipeline builder to play with, but it's a dangerous Pandora's box :)

muffato:
In the meantime I have pushed a fix for the other issues. The one related to INPUT_PLUS was in fact fixed on master, so I could simply cherry-pick the commit.

muffato:
I'd also be reluctant to extending the param-substitution language for this, and I in fact think we don't need to. But we probably need to make the following rules clear to the users:

1. Templates are always evaluated using the freshest values (i.e. the latest state of `$self->param`)
2. If the downstream analysis requires a fresh value for `alpha`, that should be explicitly told via one of these:
   - `dataflow_output_id( {'alpha' => 45} )` (no template needed)
   - `dataflow_output_id(undef)` (or the autoflow) **with a template** `{'alpha' => '#alpha#}'` which reads as: Create an input_id with a parameter named _alpha_ whose value is the freshest value of _alpha_
3. If the downstream analysis requires the original value for `alpha`, that can still be done:
   - `dataflow_output_id( {'alpha' => 23} )` (no template needed)
   - Activate `INPUT_PLUS` and: either don't use a template, or use a template that doesn't set _alpha_

If the downstream analysis also requires `'beta' => 'xx#alpha#xx'`, again there are two options:
- if we want the freshest value, `beta` should be defined in the template
- if we want the original value, `beta` should be defined in the downstream analysis itself (analysis parameter), and _alpha_ preserved with one of the above scenarios.

muffato:
I've updated the unit test based on what I've described in my previous comment

PR #27, opened 2018-04-19
Improvements to NotifyByEmail

muffato:
Addresses two comments that are only 2 years old ^^:
 - https://genomes-ebi.slack.com/archives/C0HTNH2GG/p1456929232000231
 - https://genomes-ebi.slack.com/archives/C0HTNH2GG/p1456929816000233

The NotifyByEmail Runnable was really simplistic. Switching to [Email::Stuffer](http://search.cpan.org/~rjbs/Email-Stuffer-0.014/lib/Email/Stuffer.pm) allows adding attachments very easily, and gives a better support of multi-part emails, e.g. with HTML content, etc. Note for @james-monkeyshines : Email::Stuffer is in fact just a wrapper around Email::MIME, which you are now using in EGPipeline/RNAFeatures/EmailRNAReport.pm instead of MIME::Lite, so I expect it to work in your pipelines.

MIME::Lite is not maintained any more (http://search.cpan.org/~rjbs/MIME-Lite-3.030/lib/MIME/Lite.pm) but since it was based on Email::Sender, like Email::Stuffer is, I believe collaborators account should be able to send emails too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ before_install:
 
 
 install:
-    - cpanm -v --installdeps --notest .
+    - cpanm -v --installdeps --with-recommends --notest .
     - cpanm -n Devel::Cover::Report::Coveralls
     - cpanm -n Devel::Cover::Report::Codecov
 
--- a/cpanfile
+++ b/cpanfile
@@ -24,4 +24,5 @@ recommends 'Getopt::ArgvFile';
 recommends 'BSD::Resource';
 recommends 'Chart::Gnuplot';
 recommends 'GraphViz';
+recommends 'Email::Stuffer';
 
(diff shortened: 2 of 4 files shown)


muffato:
Ah, I forgot another important change ! I have imported a piece of code originally written by @james-monkeyshines that generates a plain-text table from a matrix. My idea is to put it there so that EG pipelines don't need Common/RunnableDB/EmailReport.pm any more, but I wouldn't add extra functionalities to it as [Text::Table](http://search.cpan.org/~shlomif/Text-Table-1.133/lib/Text/Table.pm) can do the same and much more.

james-monkeyshines:
Looks good to me, thanks.

Note that for the last year or so the EG module has used Email::Sender::Simple and Email::Simple (https://github.com/EnsemblGenomes/eg-pipelines/blob/master/modules/Bio/EnsEMBL/EGPipeline/Common/RunnableDB/EmailReport.pm), so hopefully should be straightforward to retire that and standarise on the NotifyByEmail module.

PR #29, opened 2018-05-11
Show a nice error message when connecting to the database with a read-only user

muffato:
beekeeper.pl, like other scripts needs a read+write connection to the database.
As I often (by mistake) use `ensro`, I'm a bit tired of seeing this stack trace:
```
mattxps:~ $ beekeeper.pl -url 'mysql://ensro@localhost/matthieu_test_db_master'
Pipeline name: long_mult
Default meadow: LOCAL/mattxps

DBD::mysql::st execute failed: INSERT command denied to user 'ensro'@'localhost' for table 'beekeeper' at /home/matthieu/workspace/src/hive/master/modules/Bio/EnsEMBL/Hive/DBSQL/StatementHandle.pm line 119.
DBD::mysql::st execute failed: INSERT command denied to user 'ensro'@'localhost' for table 'beekeeper' at /home/matthieu/workspace/src/hive/master/modules/Bio/EnsEMBL/Hive/DBSQL/StatementHandle.pm line 119.
 at /home/matthieu/workspace/src/hive/master/modules/Bio/EnsEMBL/Hive/DBSQL/StatementHandle.pm line 145.
        Bio::EnsEMBL::Hive::DBSQL::StatementHandle::__ANON__('Bio::EnsEMBL::Hive::DBSQL::StatementHandle=HASH(0x3a61b18)', 0, 'ANALYSIS_FAILURE', 'mattxps', 'LOCAL/mattxps', 'matthieu', '-url mysql://ensro@localhost/matthieu_test_db_master', 11026, 1, ...) called at /home/matthieu/workspace/src/hive/master/modules/Bio/EnsEMBL/Hive/DBSQL/BaseAdaptor.pm line 528
        Bio::EnsEMBL::Hive::DBSQL::BaseAdaptor::class_specific_execute('Bio::EnsEMBL::Hive::DBSQL::BeekeeperAdaptor=HASH(0x3a4a680)', 'Bio::EnsEMBL::Hive::Beekeeper=HASH(0x3a4a6e0)', 'Bio::EnsEMBL::Hive::DBSQL::StatementHandle=HASH(0x3a61b18)', 'ARRAY(0x3a4e8c0)') called at /home/matthieu/workspace/src/hive/master/modules/Bio/EnsEMBL/Hive/DBSQL/BaseAdaptor.pm line 578
        Bio::EnsEMBL::Hive::DBSQL::BaseAdaptor::store('Bio::EnsEMBL::Hive::DBSQL::BeekeeperAdaptor=HASH(0x3a4a680)', 'Bio::EnsEMBL::Hive::Beekeeper=HASH(0x3a4a6e0)') called at /home/matthieu/workspace/src/hive/master/scripts/beekeeper.pl line 471
        main::register_beekeeper('Bio::EnsEMBL::Hive::Valley=HASH(0x35907a8)', 'HASH(0x23bfcb8)') called at /home/matthieu/workspace/src/hive/master/scripts/beekeeper.pl line 285
        main::main() called at /home/matthieu/workspace/src/hive/master/scripts/beekeeper.pl line 32
```
And I though I could make the error message a bit nicer.

In this commit I introduce `has_write_access` in DBConnection, which returns a boolean, and `requires_write_access`, which dies or does nothing. Each script that needs to write to the database now calls `requires_write_access`. There is one exception: beekeeper, as it still can print the status of the pipeline, a warning message about the database user and exit with the error code 1. Do you think beekeeper should just die like the other scripts ? Here is how it looks like with my changes

```
mattxps:~ $ runWorker.pl -url 'mysql://ensro@localhost/matthieu_test_db_master'
It appears that ensro doesn't have INSERT/UPDATE/DELETE privileges on this database (matthieu_test_db_master). Please check the credentials
mattxps:~ $ beekeeper.pl -url 'mysql://ensro@localhost/matthieu_test_db_master' -debug 9
Pipeline name: long_mult

**********************************************************************
* It appears that ensro doesn't have INSERT/UPDATE/DELETE privileges
* on this database (matthieu_test_db_master). Please check the credentials
*
**********************************************************************

Default meadow: LOCAL/mattxps

take_b_apart (  1)       READY, jobs(                       2r ), avg:  0.0 ms , workers(Running:0, Est.Required:1)   h.cap:-  a.cap:1  (sync'd 2789 sec ago)
part_multiply(  2)       EMPTY, jobs(                       =0 ), avg:  0.0 ms , workers(Running:0, Est.Required:0)   h.cap:-  a.cap:1  (sync'd 0 sec ago)
add_together (  3)       EMPTY, jobs(                       =0 ), avg:  0.0 ms , workers(Running:0, Est.Required:0)   h.cap:-  a.cap:1  (sync'd 0 sec ago)

total over 3 analyses :   0.00% complete (< 0.00 CPU_hrs) (2 to_do + 0 done + 0 failed + 0 excluded = 2 total)

===== Stats of active Roles as recorded in the pipeline database: ======
         ======= TOTAL ======= : 0 active Roles

**********************************************************************
* beekeeper.pl is running in read-only mode, i.e. it only
* prints the current status of the pipeline.
*
**********************************************************************
```

My other question is about the implementation of `has_write_access`. I couldn't find a clean way of checking the permissions of the current user. It looks like in theory, the permissions of a given user could vary per host it's connecting from, per database and  per table ?
I didn't feel like parsing the output of `SHOW GRANTS`, e.g.
```
| GRANT SELECT, RELOAD, SHOW DATABASES, CREATE TEMPORARY TABLES, LOCK TABLES, SHOW VIEW ON *.* TO 'ensro'@'%' |                                                                                                                                                                 
| GRANT SELECT ON `%\_%`.* TO 'ensro'@'%'                                                                     |                                                                                                                                                                 
```
and I reckon my implementation is simplistic. Is there a real risk that the user has no write permissions in general, but has write permissions enabled at the database level ?
I think it's fine not to capture all the cases where there is no write permissions (as DBI would just fail anyway when trying to write stuff) but it'd be bad preventing accessing the database if there is no real reason.
Also, I haven't searched how to check those in PostreSQL

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/DBSQL/DBConnection.pm
+++ b/modules/Bio/EnsEMBL/Hive/DBSQL/DBConnection.pm
@@ -362,5 +362,46 @@ sub run_in_transaction {
 }
 
 
+=head2 has_write_access
+
+  Example     : my $can_do = $dbc->has_write_access();
+  Description : Tells whether the underlying database connection has write access to the database
+  Returntype  : Boolean
+  Exceptions  : none
+  Caller      : general
+  Status      : Stable
+
+=cut
+
+sub has_write_access {
+    my $self = shift;
+    if ($self->driver eq 'mysql') {
+        my $a =  $self->db_handle->selectrow_arrayref('SELECT Insert_priv, Update_priv, Delete_priv FROM mysql.user WHERE user = ?', undef, $self->username);
+        return !scalar(grep {$_ eq 'N'} @$a);
+    } else {
+        # TODO: implement this for other drivers
+        return 1;
+    }
+}
+
+=head2 requires_write_access
+
+  Example     : $dbc->requires_write_access();
+  Description : See Exceptions
+  Returntype  : none
+  Exceptions  : Throws if the current user hasn't write access to the database
+  Caller      : general
+  Status      : Stable
+
+=cut
+
+sub requires_write_access {
+    my $self = shift;
+    unless ($self->has_write_access) {
+        die sprintf("It appears that %s doesn't have INSERT/UPDATE/DELETE privileges on this database (%s). Please check the credentials\n", $self->username, $self->dbname);
+    }
+}
+
+
 1;
 
--- a/modules/Bio/EnsEMBL/Hive/DBSQL/DBConnection.pm
+++ b/modules/Bio/EnsEMBL/Hive/DBSQL/DBConnection.pm
@@ -376,8 +376,12 @@ sub run_in_transaction {
 sub has_write_access {
     my $self = shift;
     if ($self->driver eq 'mysql') {
-        my $a =  $self->db_handle->selectrow_arrayref('SELECT Insert_priv, Update_priv, Delete_priv FROM mysql.user WHERE user = ?', undef, $self->username);
-        return !scalar(grep {$_ eq 'N'} @$a);
+        my $user_entries =  $self->db_handle->selectall_arrayref('SELECT Insert_priv, Update_priv, Delete_priv FROM mysql.user WHERE user = ?', undef, $self->username);
+        my $has_write_access_from_some_host = 0;
+        foreach my $entry (@$user_entries) {
+            $has_write_access_from_some_host ||= !scalar(grep {$_ eq 'N'} @$entry);
+        }
+        return $has_write_access_from_some_host;
     } else {
         # TODO: implement this for other drivers
         return 1;
--- a/modules/Bio/EnsEMBL/Hive/Scripts/InitPipeline.pm
+++ b/modules/Bio/EnsEMBL/Hive/Scripts/InitPipeline.pm
@@ -55,6 +55,7 @@ sub init_pipeline {
 
     my $hive_dba = $pipeline->hive_dba()
         or die "HivePipeline could not be created for ".$pipeconfig_object->pipeline_url();
+    $hive_dba->dbc->requires_write_access();
 
     print "> Parsing the PipeConfig file and adding objects (this may take a while).\n\n";
     $pipeconfig_object->add_objects_from_config( $pipeline );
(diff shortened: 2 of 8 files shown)


ens-bwalts:
Beekeeper death: I can see a case for the beekeeper to keep looping (user wants to monitor a pipeline or service from somewhere, but doesn't want to submit workers or accidentally mess anything up), so I think this implementation works.

As for how has_write_access should be implemented:
Not sure about checking db-level access. However, consider changing the implementation to use selectall_arrayref and then if any row for the user has INSERT + UPDATE + DELETE return true, to get around the problem of having different permissions from different hosts. This is a bit of an overgeneralization, but I'm not sure that it's worth trying to reliably match the hostname that the script thinks it's running on to the hostname that the mysql server sees.

e.g.
```
my $user_entries =  $self->db_handle->selectall_arrayref('SELECT Insert_priv, Update_priv, Delete_priv FROM mysql.user WHERE user = ?', undef, $self->username);
my $has_write_access_from_some_host = 0
foreach my $entry (@$user_entries) {
     $has_write_access_from_some_host = $has_write_access_from_some_host || 
         !scalar(grep {$_ eq 'N'} @$entry);
}
return $has_write_access_from_some_host;
```

muffato:
Good point, it could be useful to keep looping in read-only mode to monitor a pipeline without messing anything up. Currently I avoid this by setting `max_loops` to 0, which then skips `run_autonomously`, but I'll try to fix the latter

muffato:
Regarding using `selectall` for the detection of write access, that's a good idea. Thanks ! And yes I agree, fully matching hostnames (IP, name, aliases, domain name, etc) is not worth doing here

muffato:
beekeeper can now loop with a read-only user, but it will only show the stats. As it's got a read-only access to the db:
 - it won't register itself (no beekeeper_id)
 - it won't sync any analyses: it relies on the workers / an other beekeeper to keep the stats up to date
 - obviously, it won't submit anything, but it won't print any intentions either. This is because the script currently calls `schedule_workers_resync_if_necessary` which may want to re-sync the analyses, and I didn't want to make too many changes

PR #35, opened 2018-05-25
bugfix: the / is a remnant of the previous regex

muffato:
This has already been fixed on master (via 5a40ea190) and I'm backporting it to 2.4


muffato:
Sorry, should have been clearer. Because of the trailing /, eHive was not checking whether the analysis exists or not

PR #36, opened 2018-05-25
Add sections about resource usage, optimization, monitoring, and update the changelog

ens-bwalts:


docs/appendix/changelog.rst
```diff
@@ -1,6 +1,82 @@
...
+* runWorker.pl
+    * runWorker.pl can now be passed a specific ``HiveConfig.json`` with the ``--config_file`` option.
+* visualize_jobs.pl
+    * New visualize_jobs.pl script which shows a Job-level picture of an eHive pipeline, including parameter and Accumulator values.
+
+Database related updates
+------------------------
+
+* Database handles now protected from timed-out connections.
+* More flexible quoting rules for URL syntax.
```
muffato: I'm not sure what this means ! What is now allowed that was not before ?

muffato:
Another feature that's worth mentioning: there is now a `hive_default_max_retry_count` PipeConfig parameter to change the default value (3) for _all_ analyses in one go.

muffato:
Also, the nested bullet lists look a bit unusual, RTD writes the outer list in bold face, and insert a some <dl><dt> elements in the HTML. I wonder if it'd need some extra blank lines ?

docs/appendix/changelog.rst
ens-bwalts: There were issues where quoted URLs would sometimes not get parsed correctly, which conflicted with the fact that you had to quote a URL if it had a semicolon in it on the command line (to protect it from the shell). That's fixed now, so you can single or double quote URLs wherever you want and they get passed around correctly.

muffato: OK, then I'd add "allowing extra parameters to be passed"

PR #40, opened 2018-07-26
enable overflow for accu keys, to_analysis_urls, and submission_cmd_args

ens-bwalts:


modules/Bio/EnsEMBL/Hive/DBSQL/ResourceDescriptionAdaptor.pm
```diff
@@ -54,6 +54,13 @@ sub default_insertion_method {
 }
 
 
+sub overflow_limit {
+    return {
+        'submission_cmd_args' => 255,
```
muffato: I would add `worker_cmd_args` because of the `reg_conf`, `hive_log_dir` and `worker_log_dir` topions

modules/Bio/EnsEMBL/Hive/DBSQL/AccumulatorAdaptor.pm
```diff
@@ -47,6 +47,32 @@ sub default_table_name {
...
+                   }
+
+
+             } @$fields ];
+}
+
+sub overflow_limit {
+    return {
+        'key_signature' => 255,
+        'struct_name'   => 255,
```
muffato: Not sure about overflowing `struct_name`. I see it as a static identifier / parameter name, like `logic_name`, `pipeline_wide_parameters.param_name`, etc. I also think they should be protected, but perhaps differently. Maybe introduce a `size_limit` method in the adaptors that define a maximum allowed size for some fields, and the base `store` method checks the values before doing the INSERT ? It might be possible to get those limits automatically through `DBSQL::BaseAdaptor::_table_info_loader`

modules/Bio/EnsEMBL/Hive/DBSQL/AccumulatorAdaptor.pm
```diff
@@ -62,6 +88,9 @@ sub fetch_structures_for_job_ids {
 
         ROW: while(my ($receiving_job_id, $struct_name, $key_signature, $stringified_value) = $sth->fetchrow_array() ) {
 
+            ($key_signature, $struct_name) = map {$self->check_and_dereference_analysis_data($_)}
+                ($key_signature, $struct_name);
+
```
muffato: Do you need to "manually" call `check_and_dereference_analysis_data` here because it's the only place that does direct SQL access, whereas all the other adaptors naturally get the overflow working through fetch_all/slicer ?
Good catch !

modules/Bio/EnsEMBL/Hive/DBSQL/AccumulatorAdaptor.pm
```diff
@@ -47,6 +47,32 @@ sub default_table_name {
...
+                          if (defined($ol) and defined($value) and  (length($value) > $ol)) {
+                              $self->db->get_AnalysisDataAdaptor()->store_if_needed($value);
+                          } else {
+                              $value;
+                          }
+                   }
+
+
+             } @$fields ];
+}
```
muffato: Did you override `slicer` here because AccumulatorAdaptor is the only NakedTableAdaptor that has overflowing and you wanted to keep NakedTableAdaptor's `slicer` simple and straightforward ?

It just felt surprising at the first glance. Maybe you can add a comment on top of NakedTableAdaptor::slicer indicating that should we need to move the overflow capability up, we can find it in AccumulatorAdaptor ?

modules/Bio/EnsEMBL/Hive/DBSQL/AccumulatorAdaptor.pm
ens-bwalts: The other exception is in AnalysisJobAdaptor, which also does this check when it queries directly rather than delegating up to BaseAdaptor::fetch_all - which also should be updated to use check_and_dereference_analysis_data now that it's available.

modules/Bio/EnsEMBL/Hive/DBSQL/AccumulatorAdaptor.pm
ens-bwalts: Yes - I didn't want to impose the overhead on every inheriting Adaptor. However, the overhead is small, and for maintainability it's probably better to move it to the base class

modules/Bio/EnsEMBL/Hive/DBSQL/AccumulatorAdaptor.pm
ens-bwalts: struct_name needs protection because it's linked to to_analysis_url, which is itself protected. As for param_name, logic_name, and others that get set at init time - I'm not sure it's worth introducing a size limit which would throw an error, since the MySQL error that gets thrown is actually informative in this case.

muffato: It actually depends how the MySQL server has been configured. mysql-ens-compara-prod-1 complains:
> DBD::mysql::st execute failed: Data too long for column 'logic_name' at row 1 at /home/matthieu/workspace/src/hive/2.5/modules/Bio/EnsEMBL/Hive/DBSQL/StatementHandle.pm line 140.

whereas mysql-ens-reg-prod-4 silently truncates the logic name to 256 characters

modules/Bio/EnsEMBL/Hive/DBSQL/NakedTableAdaptor.pm
```diff
@@ -47,7 +47,18 @@ use base ('Bio::EnsEMBL::Hive::DBSQL::BaseAdaptor');
 sub slicer {    # take a slice of the hashref (if only we could inline in Perl!)
     my ($self, $hashref, $fields) = @_;
 
-    return [ @$hashref{@$fields} ];
+    my $overflow_limit = $self->overflow_limit();
+
+    return [ map { eval { my $value = $hashref->{$_};
```
muffato: I know there is an `eval` in ObjectAdaptor too, but do we really need them ? I am concerned they could hide some errors

modules/Bio/EnsEMBL/Hive/DBSQL/NakedTableAdaptor.pm
```diff
@@ -47,7 +47,18 @@ use base ('Bio::EnsEMBL::Hive::DBSQL::BaseAdaptor');
...
+                          my $ol = $overflow_limit->{$_};
```
muffato: This is almost exactly the same as in ObjectAdaptor. I would prefer this to be in BaseAdaptor but I'm not sure how to do it cleanly, clearly and efficiently.

modules/Bio/EnsEMBL/Hive/DBSQL/AccumulatorAdaptor.pm
muffato: Remove `struct_name` ?

ens-bwalts: Have been experimenting (not pushed yet) with getting sizes in _table_info_loader(). Straightforward, but for many types COLUMN_SIZE returns null. These are types such as INT or TEXT, where either the size is very large, or that we are less concerned about. The bottom line being that the check against size_limit is not comprehensive without a lot of work (and somehow defining it for every column elsewhere in the code), but it does cover the cases that the ticket and this PR care about

modules/Bio/EnsEMBL/Hive/DBSQL/BaseAdaptor.pm
```diff
@@ -229,18 +250,27 @@ sub _table_info_loader {
...
         # warn "ColumnInfo [$table_name/$column_name] = $column_type\n";
 
         $column_set{$column_name}  = $column_type;
 
+        # PostgreSQL reports a COLUMN_SIZE of 4 for enums, which is not compatible with
+        # the way slicer does column size checking. Likewise, PostgreSQL reports
+        # a user-defined TYPE_NAME for enums, rather than 'enum'. Therefore, if
+        # the DB is PostgreSQL, only set size_limit for varchars
+        unless (($driver eq 'pgsql') && !($column_type eq 'character varying')) {
+            $size_limit{$column_name}  = $size_limit;
```
muffato: This will set a size limit for _all_ column types, incl. integers, enums and timestamps, but I'm not sure these are correctly enforceable:
- For integers, we get the size of the longest possible string that fits in. For instance, for the standard INT (32 bits) which ranges from -2^32 to 2^32-1, that's 11 characters, but in fact, not all 11-digit integers can fit there.
- ENUMs are given the size of their longest element
- TIMESTAMPs are given 14 characters, which I find scary. My text editor tells me that `2019-04-05 15:47:14` has 19 characters, so the API would start throwing when storing timestamps ?? Although these columns are most often set to CURRENT_TIMESTAMP in a way that bypasses `slicer`, I fear a corner case whereby eHive stores a previously-fetched object that already has a timestamp and the adaptor would now fail.

I would suggest to only populate `size_limit` for text fields.

modules/Bio/EnsEMBL/Hive/DBSQL/AccumulatorAdaptor.pm
```diff
@@ -48,6 +48,13 @@ sub default_table_name {
 }
 
 
+sub overflow_limit {
```
nerdstrike: It's odd to me how many times you overload the overflow_limit sub.

EbiArnie:
This hasn't been urgently requested in five years, so I will go ahead and close this without merging.

PR #41, opened 2018-07-26
Expose DBI::st's hash interface through StatementHandle

muffato:
Applications need to access DBI::st hash values (like $sth->{Active}) and
the recent change in StatementHandle broke those.
Here we reenable the hash interface by redirecting all the calls to the
statement handle.

In the end, I think this is the best solution as it provides backwards compatibility for all existing code. In fact, you can notice that the AUTOLOAD was basically already redirecting all the method calls to the DBI::st instance, so it makes sense redirecting the hash methods too !

I have expanded a unit-test to also check some of those sth variables (`Active` and `NAME`) and they work !

I think I could then close Ensembl/ensembl#270 ?

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/DBSQL/StatementHandle.pm
+++ b/modules/Bio/EnsEMBL/Hive/DBSQL/StatementHandle.pm
@@ -58,6 +58,12 @@ sub new {
 
     my $self = bless {}, $class;
 
+    my $real_self = {};
+    # $self will remain empty whereas $real_self will actually have all the data
+    # It's only purpose is to offer a hash-reference on which perl allows
+    # calling hash accessors, e.g. $sth->{Active}
+    tie %$self, 'DBIstHashProxy', $real_self;
+
     $self->dbc( $dbc );
     $self->sql( $sql );
     $self->attr( $attr );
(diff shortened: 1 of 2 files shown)


muffato:
See #42 for an alternative implementation. #42 is in fact the first approach I tried, but I got a bit lost trying to understand how tie-ing works, and ended up with this one here. But now it's much clearer in my mind and I could make `Tie::ExtraHash` work too (it actually perfectly fits cases like ours)

magaliruffier:
Don't you love how Perl has multiple solutions for the same problem? ;-)
#42 looks neater, less code to update if anything changes

modules/Bio/EnsEMBL/Hive/DBSQL/StatementHandle.pm
```diff
@@ -179,4 +198,29 @@ sub DESTROY {   # note AUTOLOAD/DESTROY interdependence!
...
+sub TIEHASH {
+    my $class = shift;
+    my $self = $_[0];
+    bless $self, $class;
+    return $self;
+}
+
+sub STORE    { $_[0]->{'_dbi_sth'}->{$_[1]} = $_[2] }
+sub FETCH    { $_[0]->{'_dbi_sth'}->{$_[1]} }
+sub FIRSTKEY { my $a = scalar keys %{$_[0]->{'_dbi_sth'}}; each %{$_[0]->{'_dbi_sth'}} }
```
nerdstrike: What does $a do here?

muffato: Actually, I don't really know. I've copied it from the default implementation of Tie::StdHash, see /nfs/software/ensembl/RHEL7-JUL2017-core2/plenv/versions/5.24.1/lib/perl5/5.24.1/Tie/Hash.pm lines 249-256 (I can't find it online !)
My guess is is that's it's there to reset the internal cursor.

muffato:
Yes it'd be useful to discuss. eHive's DBConnection has diverged because it was not possible to enhance Core's DBConnection, but there are certainly a few things that could be ported back

PR #42, opened 2018-07-26
Expose DBI::st's hash interface through StatementHandle

muffato:
See #41 for the purpose of the change
This is an alternative implementation.


ens-bwalts:
Agreed that this solution is simpler - and better to farm out implementation to Tie::Extrahash.

PR #44, opened 2018-08-14
Minor improvement to the resource usage view

muffato:
This PR won't pass as the changes should be accompanied with SQL patches, but I wanted you to have a look at it first. If you have any suggestions about the views, or feedback from people, it'd be great to bundle all the changes.

For instance, I renamed `count(*)` to `num_workers` but I've now realised that a similar column in `resource_usage_stats` is called `workers`, whereas in `progress` it's just `cnt`

Matthieu

### DIFF
--- a/sql/procedures.mysql
+++ b/sql/procedures.mysql
@@ -74,9 +74,9 @@ CREATE OR REPLACE VIEW resource_usage_stats AS
            w.meadow_type,
            CONCAT(rc.name,'(',rc.resource_class_id,')') resource_class,
            u.exit_status,
-           count(*) workers,
+           count(*) AS num_workers,
            min(mem_megs) AS min_mem_megs, round(avg(mem_megs),2) AS avg_mem_megs, max(mem_megs) AS max_mem_megs,
-           min(swap_megs) AS min_swap_megs, round(avg(swap_megs),2) AS avg_swap_megs, max(swap_megs) AS max_swap_megs
+           round(min(cpu_sec/lifespan_sec),2) AS min_cpu_usage, round(avg(cpu_sec/lifespan_sec),2) AS avg_cpu_usage, round(max(cpu_sec/lifespan_sec),2) AS max_cpu_usage
     FROM resource_class rc
     JOIN analysis_base a USING(resource_class_id)
     LEFT JOIN role r USING(analysis_id)
@@ -93,7 +93,7 @@ CREATE OR REPLACE VIEW resource_usage_stats AS
 --       select * from live_roles where resource_class_id=12;
 
 CREATE OR REPLACE VIEW live_roles AS
-    SELECT w.meadow_user, w.meadow_type, w.resource_class_id, rc.name resource_class_name, r.analysis_id, a.logic_name, count(*)
+    SELECT w.meadow_user, w.meadow_type, w.resource_class_id, rc.name resource_class_name, r.analysis_id, a.logic_name, count(*) AS num_workers
     FROM worker w
     JOIN role r USING(worker_id)
     LEFT JOIN resource_class rc ON w.resource_class_id=rc.resource_class_id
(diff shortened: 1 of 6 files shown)


muffato:
Maybe @ens-bwalts you could add your semaphore view to this branch and we'll squash everything as a single schema change ?

sql/procedures.mysql
```diff
@@ -122,6 +122,55 @@ CREATE OR REPLACE VIEW beekeeper_activity AS
...
+    LEFT JOIN analysis_base s_ab ON s_ab.analysis_id = s_j.analysis_id;
+
+-- show a summary of semaphored jobs and/or dependent semaphore URLs
+--
+-- Usage:
+--       select * from semaphore_sumary;
+--       select * from semaphore_summary where semaphored_logic_name like '%nalysi%';
+
+CREATE OR REPLACE VIEW semaphore_summary AS
+   SELECT s.semaphore_id, s_ab.logic_name as semaphored_logic_name, s.dependent_job_id as semaphored_job_id,
```
muffato: This causes an error (I think only under PostgreSQL at the moment, but recent versions of MySQL also check this)
```
ERROR:  column "s.semaphore_id" must appear in the GROUP BY clause or be used in an aggregate function
LINE 2:    SELECT s.semaphore_id, s_ab.logic_name as semaphored_logi...
```
I think you can use `s.semaphore_id` in the GROUP BY ?

sql/procedures.mysql
```diff
@@ -122,6 +122,55 @@ CREATE OR REPLACE VIEW beekeeper_activity AS
...
+   NULL as dependent_semaphore_url, c_ab.logic_name as controlling_logic_name, count(*) as job_count,
```
muffato: I wonder if you forgot to add `c_j.analysis_id` in the GROUP BY so that it shows the job count for each pair of analysis. Otherwise I'd rename this column to `example_controlling_logic_name`

sql/procedures.mysql
```diff
@@ -122,6 +122,55 @@ CREATE OR REPLACE VIEW beekeeper_activity AS
...
+--
+-- Usage:
+--       select * from job_semaphores;
+--       select * from job_semaphores where controlling_logic_name = 'an_analysis';
+--
+--       -- get semaphore counts for jobs of a particular semaphored analysis
+--       select distinct semaphored_job_id, semaphore_count from job_semaphores
+--       where semaphored_logic_name = 'an_analysis'
+
+CREATE OR REPLACE VIEW job_semaphores AS
```
muffato: Need to add both views to the list in `drop_hive_tables`

ens-bwalts:
Removed the semaphores_by_factory view. It is very complex to do as a view - the difficulty is knowing where to stop when traversing prev_job_id to find the actual factory. It may be possible, since you can get the analysis of the factory job by going through the dataflow_rules table. Finding jobs by factory may work better as a stored procedure.

muffato:
Is the factory not the parent of the funnel ?

I would build the view by linking the semaphore table to the job table (for the funnel), the job table again to get the funnel's parent (the factory), and another time to get all the jobs that belong to the semaphore

muffato:
Regarding  `semaphore_summary`:
I would suggest dropping `remote_jobs_counter` because it gives the wrong impression it is the number of remote jobs associated with the `controlling_logic_name` analysis, and also the same value would be repeated if there are several analyses controlling the semaphore. Maybe also rename `fan_jobs_this_analysis` to `remaining_jobs_this_analysis` to indicate that it's only what's left to run

muffato:
@ens-bwalts : Since you've said you're happy with #64, I will rebase the branch and remove your commits. I'll also check that the names are consistent (e.g. `num` vs `cnt` vs `count`).

Would we fuse both branches in order to have just 1 schema version change ?

ens-bwalts:
Yes, the fewer schema version changes the better

PR #45, opened 2018-08-14
Experimental/misc master

muffato:
Here is a list of minor changes I've made whilst working on the scheduler. Users will notice that beekeeper's output looks different, e.g.

```
mattxps:~ $ beekeeper.pl -url mysql://ensadmin:ensembl@localhost:3306/matthieu_long_mult --debug 1 -analyses_pattern 2,3,1
Pipeline name: long_mult
Default meadow: LOCAL/mattxps

Beekeeper : the following Analyses matched your -analyses_pattern '2,3,1' : take_b_apart(1), part_multiply(2), add_together(3)
Beekeeper : 0 Analyses are not shown

 take_b_apart(1)   DONE,           2002d jobs,  avg: 33.2 sec,  0 workers ( 0 required),  h.cap:- a.cap:-  (sync'd 40119 sec ago)
part_multiply(2)  EMPTY,              =0 jobs,  avg: N/A,       0 workers ( 0 required),  h.cap:- a.cap:-  (sync'd 40116 sec ago)
 add_together(3)  READY,  95r+1555d=1650 jobs,  avg: N/A,       0 workers (95 required),  h.cap:- a.cap:-  (sync'd   747 sec ago)

total over 3 analyses :  97.40% complete (< 0.00 CPU_hrs) (95 to_do + 3557 done + 0 failed + 0 excluded = 3652 total)

===== Stats of active Roles as recorded in the pipeline database: ======
         ======= TOTAL ======= : 0 active Roles

Scheduler : Discarded 2 analyses because they do not need any Workers.
Scheduler : Analysis 'add_together' is READY, safe-synching it... succeeded
Scheduler : add_together(3)  READY,  95r+1555d=1650 jobs,  avg: N/A,   0 workers (95 required),  h.cap:- a.cap:-  (sync'd 0 sec ago)
Scheduler : Hit the limit of *** Max number of Workers scheduled this time ***, settling for 50 Workers.
Scheduler : Hit the limit of *** Number of workers in 'LOCAL/mattxps' meadow ***, settling for 4 Workers.
Scheduler : The Scheduler allocated 4 x LOCAL:default extra workers for 'add_together' [1.0000 hive_load remaining]
Scheduler : I recommend submitting 4 x LOCAL:default workers
```
instead of
```
mattxps:~ $ beekeeper.pl -url mysql://ensadmin:ensembl@localhost:3306/matthieu_long_mult --debug 1 -analyses_pattern 2,3,1
Pipeline name: long_mult
[...]

PR #46, opened 2018-08-14
Improved scheduler

muffato:
## Main changes

Following #45, here is the meat of the change. There are three major new features

1. Automatic setting of the batch-size.

in fact, the feature has been there for a very very long time. I just wasn't very happy with the formula. I now try to achieve 5 claim operations per analysis per second. The estimated batch size is simply the optimal number of jobs to achieve that given the number of running workers and the average runtime of a job.

2. New limiter based on the job throughput

This is similar to `hive_capacity` in the sense that all the analyses will be counted in the same limiter and limit each other, but the difference is that the average runtime of jobs is taken into account. The limiter caps the total _throughput_ (jobs per second) of the pipeline. I've checked most of the Compara databases of the last 6 months or so, and the maximum we could reach is ~2,000 jobs per second. However I will ask feedback on #ensembl-hive

3. Submit fewer workers and ensure they all have enough work to do

The scheduler was very greedy and was basically trying to submit 1 worker per job (before limiting this number according to the capacities). For instance, if there are 20 jobs to run, and we know that the average job runtime is 1 second, eHive will still submit 20 workers even though we could afford submitting a single one. Submitting fewer workers is nicer to LSF, especially if you consider that very short jobs are penalized (the `underrun` exception), and that because workers are not guaranteed to start at the same time, when the 20th worker starts, the analysis may have been depleted.
The scheduler now aims at feeding each worker with ~2 minutes worth of work. There is a mechanism similar to the TailTrimming so that it will still submit a few more workers at the next loop, which helps when job runtimes are heterogeneous and workers get stuck on long jobs.

## Related changes

1. I've made `estimate_num_required_workers` return the number of extra workers needed. In fact, it was used like that in Scheduler although defined as the total number of workers in AnalysisStats, but was working fine only because the limiters were enforced in both.

2. I have added a method in AnalysisStats to estimate `avg_msec_per_job` when no workers have reported their stats yet. This is done by reading how many live workers we have, and how many jobs they've been involved with. Even though MySQL timestamps only store whole seconds, it is still sufficient in most cases as the new mechanisms are mostly effective on very quick analyses

## Other comments

This work is largely motivated by the issues encountered by someone new to eHive and trying to run a pipeline on a large dataset (e.g. LongMultiplication with tens of thousands of jobs). Ideally, people should not bother about batch-size, capacity, and eHive should run sensibly (I guess you'll remember which presentation I'm alluding to).

## Things left to do and test before this PR can be accepted

1. I've tested the automatic batch-size on several large Compara pipelines and it worked well. I have only tested the other changes in small pipelines, but will test them at scale too.

2. By default, the batch-size is set to 1 in Analysis.pm (unless defined otherwise). I will run a few more tests, but I would in fact change that to 0 so that the automatic setting is used and we can tell people to forget having to set batch-size

3. I've defined all the parameters as constants in AnalysisStats, but I feel they should ultimately go to the JSON config file. I think we've already mentioned we should properly version it and use it adequately, and I guess it will become more important now

4. I will need to update the documentation

5. Maybe I should include those explanations in the commit messages themselves ?

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
+++ b/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
@@ -312,10 +312,13 @@ sub estimate_num_required_workers {     # this doesn't count the workers that ar
     # 1) hive_capacity
     my $h_cap = $self->analysis->hive_capacity;
     if( defined($h_cap) and $h_cap>=0) {  # what is the currently attainable maximum defined via hive_capacity?
+        # NOTE: $hive_current_load can be higher than 1, making $h_max lower than 0
         my $hive_current_load = $self->hive_pipeline->get_cached_hive_current_load();
         # Round to 3 places before taking the integral part
         my $h_max = POSIX::floor( sprintf('%.3f', $h_cap * ( 1.0 - $hive_current_load )) );
-        if($h_max < $num_required_workers) {
+        if ($h_max <= 0) {
+            $num_required_workers = 0;
+        } elsif ($h_max < $num_required_workers) {
             $num_required_workers = $h_max;
         }
     }
--- a/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
+++ b/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
@@ -269,8 +269,38 @@ sub estimate_num_required_workers {     # this doesn't count the workers that ar
     my $self                = shift @_;
     my $remaining_job_count = shift @_ || 0;    # FIXME: a better estimate would be $self->claimed_job_count when it is introduced
 
-    my $num_required_workers = $self->ready_job_count + $remaining_job_count;   # this 'max' estimation can still be zero
+    # Work left to do
+    my $jobs_to_do = $self->ready_job_count + $remaining_job_count;
 
+    # We can assume jobs_to_do>0 i the rest of the function
+    return 0 unless $jobs_to_do;
+
+    # Guaranteed to be non-zero
+    my $avg_msec_per_job = $self->get_or_estimate_avg_msec_per_job;
+    # Number of jobs earmarked for each worker, existing or new. (cannot be zero)
+    # NOTE: With the current defaults, this is 12,000 when the analysis has never run
+    my $num_jobs_per_worker = POSIX::ceil($self->min_worker_runtime / $avg_msec_per_job );
+
+    # Total number of jobs earmarked for the running workers (knowing they are already running 1 job each)
+    # NOTE: This is only greater than zero when jobs are quicker than $self->min_worker_runtime
+    my $jobs_for_running_workers = $self->num_running_workers * ($num_jobs_per_worker - 1);
+
+    # NOTE: This means 1 worker per job when jobs_for_running_workers = 0 (no running workers or jobs take more than min_worker_runtime)
+    my $num_required_workers = POSIX::ceil( ($jobs_to_do - $jobs_for_running_workers) / $num_jobs_per_worker);
+    # NOTE: Since it can be negative when there are fewer jobs to do than what we have earmarked for running workers,
+    #       we apply a correction and always submit at least 1 worker (since jobs_to_do is > 0)
+    $num_required_workers = 1 if $num_required_workers < 1;
+
+    ## Alternatively
+    #if ($jobs_to_do > $jobs_for_running_workers) {
+    #    # More jobs than what the existing workers can achieve
+    #    $num_required_workers = POSIX::ceil( ($jobs_to_do - $jobs_for_running_workers / 2) / $num_jobs_per_worker);
+    #} else {
+    #    # In principle all jobs could be done by existing workers, but we still submit some workers in case some workers are stuck
+    #    $num_required_workers = POSIX::ceil( ($jobs_to_do / 2) / $num_jobs_per_worker);
+    #}
+
+    # 1) hive_capacity
     my $h_cap = $self->analysis->hive_capacity;
     if( defined($h_cap) and $h_cap>=0) {  # what is the currently attainable maximum defined via hive_capacity?
         my $hive_current_load = $self->hive_pipeline->get_cached_hive_current_load();
@@ -279,6 +309,8 @@ sub estimate_num_required_workers {     # this doesn't count the workers that ar
             $num_required_workers = $h_max;
         }
     }
+
+    # 2) analysis_capacity
     my $a_max = $self->analysis->analysis_capacity;
     if( defined($a_max) and $a_max>=0 ) {   # what is the currently attainable maximum defined via analysis_capacity?
         # Correction for running workers
--- a/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
+++ b/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
@@ -290,6 +290,12 @@ sub estimate_num_required_workers {     # this doesn't count the workers that ar
         }
     }
 
+    # 3) job throughput
+    my $t_max = POSIX::floor( $avg_msec_per_job * ($self->max_jobs_per_msec - $self->hive_pipeline->get_total_job_throughput) );
+    if($t_max < $num_required_workers) {
+        $num_required_workers = $t_max;
+    }
+
     return $num_required_workers;
 }
 
--- a/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
+++ b/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
@@ -77,6 +77,11 @@ sub min_job_runtime_msec {
     return 10;
 }
 
+# How long the stats are valid or. Used in synchronize_AnalysisStats
+sub stats_expiration_date_sec {
+    return 3*60;
+}
+
 
 =head1 AUTOLOADED
 
--- a/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
+++ b/modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
@@ -290,10 +290,43 @@ sub estimate_num_required_workers {     # this 'max allowed' total includes the 
 }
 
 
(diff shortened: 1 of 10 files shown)


modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
```diff
@@ -249,31 +265,99 @@ sub get_or_estimate_batch_size {
...
+    # Work left to do
+    my $jobs_to_do = $self->ready_job_count + $remaining_job_count;
+
+    # Correction so that 1) each worker runs for a minimum amount of time
+    # and 1) existing workers get a chance to do some more work besides
+    # their current job.
+    # NOTE: the /2 factor is the equivalent of the TailTrimming mechanism
+    # NOTE: if the analysis has never been run, it will assign
+    #       min_worker_runtime/min_job_runtime_msec=12,000 jobs to each worker
+    my $num_required_workers = POSIX::ceil( ($jobs_to_do + $self->num_running_workers) / $num_jobs_per_worker - $self->num_running_workers/2 );
```
muffato: This is not the right formula yet. It doesn't obey the constraints:

* `$jobs_to_do` == 0 -> `$num_required_workers` = 0 (regardless of `$self->num_running_workers`), i.e. stop submitting workers when there is nothing left to do
* `$num_required_workers` <= `$jobs_to_do`, i.e don't submit more workers than there are jobs to do

ens-lg4:
Wow, this looks like a very exciting change - fulfilling the actual point of having a scheduler and collecting all the job stats, which was to leave the scheduling details to automation and let the users concentrate on their application-specific code.

Is this change backwards-compatible with old style "chatty" config files that explicitly specify all the capacities and batch_sizes?

muffato:
Thank you @ens-lg4 ! Yes it still understands user-defined capacities and batch-size. Capacities will still be needed until eHive has a way of assessing the load on all the resources the analyses need (not sure how to get that done anyway), but I hope people will be able to stop having to set batch_size

muffato:
I have fixed the formula for `num_required_workers`, it now honours the constraints I've laid out in my other comment. I have come up with two ways of estimating the number of workers.

In one formula, the scheduler earmarks some jobs for the running workers (as long as the average runtime is not too high) and only submits new workers for the other jobs. However it always submits 1 worker in case an existing worker gets stuck.
In the other formula, the scheduler earmarks some jobs for running workers the same way but submits new workers for both the remaining jobs and half of the running workers. It doesn't rely as much on existing workers as the first formula.

I will try both on Compara pipelines. My objective is to stop seeing workers quitting immediately because there is no job for them, decreasing the total number of workers (thus increasing the average number of jobs per worker) whilst not increasing the runtime of the pipeline, especially the tail of each analysis.

muffato:
I haven't done as many tests as I was hoping to but here is what I have so far.

I have compared a configuration of the Compara Protein-trees pipeline (~600K jobs and ~36h wall-time) between the new code and v2.5. With the new scheduler, the pipeline can complete with ~12K workers instead of ~18K, each worker running ~48 min (51 jobs) instead of 30 min (35 jobs). 200 workers can't find any role instead of 1,100, and more generally, 350 workers last for less than 5 seconds instead of 1,400. Surprisingly, only 1,500 (12%) of the workers respecialised, instead of 2,900 (16%). The total wall-time seemed unaffected.

However I have spotted few other scheduler issues that I want to address too:

1) In another pipeline, I have an analysis with a high memory requirement and a high hive priority. LSF could not start all the workers and I quickly reached 50 pending workers. At that point, beekeeper was basically just monitoring those 50 pending workers, adding 1 as soon as 1 could start running. There were very few running workers (of this analysis) and I feel like beekeeper could have submitted workers of the other analyses in the meantime. I will try to change the correction that is being applied. Something along the lines of: workers that haven't pending long enough are not expected to start soon and are not considered by the scheduler. The scheduler can use those slots to schedule other analyses instead. It would break the current rule "workers are not allowed to specialize in an analysis if an analysis of higher priority needs workers", though

2) LOADING analyses have a lower internal priority than READY or WORKING analyses, because they require a sync (according to code comments). In a pipeline I've got a _lomem analysis that spills in an _himem analysis (which has a higher hive priority). Because of the regular flow of jobs, the status of the _himem analysis is often LOADING, meaning that the scheduler will prefer other analyses even if they have a lower priority. This in fact already breaks the rule I've mentioned in my previous point.

3) When an analysis is generally failing because of say a missing parameter, beekeeper basically keeps on submitting 50 workers every loop, increasing the retry_count of 100 jobs each time. Under the default settings it takes 3*ceiling(n_jobs/100) minutes to see the first failure. With the new scheduler, assuming the error arises quickly and the jobs are fast, beekeeper may submit less than 50 workers each time, making this even slower. For example, a 50 ms job average means beekeeper earmarks 2,400 jobs per workers, so when there are less than 2,400 jobs to do, only 1 worker is submitted every loop and it takes 40 hours to try each job once instead of 24 min

muffato:
There are a few more things I would like to push to this branch before merging it. Especially:
 - The changes here make beekeeper submit fewer workers based on runtime stats. I've found a situation where it submits just 1 worker because it thinks the analysis is very fast, but the remaining jobs are very slow, and thus it takes ages to get it done. I think the scheduler was too "aggressive" and now is too "parsimonious". I would like to find a better balance.
 - I feel like some of the changes should be considered as bugs and backported to earlier branches, cf the situation that Carla reported on Slack last month: https://genomes-ebi.slack.com/archives/C0F9K0G2E/p1550238716061700

GitHub have recently introduced the concept of "draft" pull-requests for PRs but I can't find a way of turning this one into a draft, I think it's only for new PRs that you can chose ... But I very much consider this PR as a draft

modules/Bio/EnsEMBL/Hive/AnalysisStats.pm
```diff
@@ -249,31 +276,129 @@ sub get_or_estimate_batch_size {
...
+    my $self = shift;
+
+    return unless $self->num_running_workers;
+
+    # No role has yet been finalized otherwise avg_msec_per_job would be set
+    my $active_roles = $self->adaptor->db->get_RoleAdaptor->fetch_all( 'when_finished IS NULL AND analysis_id=' . $self->analysis_id );
+    return unless @$active_roles;
+
+    my $tot_lifespan_secs = sum( map {$_->seconds_since_when_started} @$active_roles );
+    return unless $tot_lifespan_secs;;
```
mira13: Please remove second ;

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisDataAdaptor.pm
```diff
@@ -81,7 +81,8 @@ sub store_if_needed {
     # UNIQUE KEY in the table definition.
     # This is very similar to check_object_present_in_db_by_content()
     # but it returns the *first* analysis_data_id that's been stored
-    my $sql = 'SELECT MIN(analysis_data_id) FROM analysis_data WHERE md5sum = ? AND data = ?';
+    my $table_name = $self->table_name();
+    my $sql = "SELECT MIN(analysis_data_id) FROM $table_name WHERE md5sum = ? AND data = ?";
```
mira13: Is it possible that we failed to obtain table name, so request will be failed and we don't process that case, probable need to add error on execute fail?

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisStatsAdaptor.pm
```diff
@@ -193,5 +193,18 @@ sub increment_a_counter {
...
+
+    my $sql = 'SELECT '  . $self->dbc->_interval_seconds_sql('MAX(when_logged)') . ' FROM analysis_stats_monitor WHERE analysis_id = ? AND sync_lock = 0';
+    my $sth = $self->prepare($sql);
+    $sth->execute($analysis_id);
+    my ($seconds_since_locked) = $sth->fetchrow_array();
+    $sth->finish;
+    return $seconds_since_locked;
+}
+
+
```
mira13: The same, do we need process case when request is failed?

modules/Bio/EnsEMBL/Hive/DBSQL/BeekeeperAdaptor.pm
```diff
@@ -117,7 +117,8 @@ sub bury_other_beekeepers {
 sub reload_beekeeper_is_blocked {
     my ($self, $beekeeper) = @_;
 
-    my $query = 'SELECT is_blocked FROM beekeeper WHERE beekeeper_id = ?';
+    my $table_name = $self->table_name();
+    my $query = "SELECT is_blocked FROM $table_name WHERE beekeeper_id = ?";
```
mira13: Same comment

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisDataAdaptor.pm
muffato: This commit (using `$self->table_name()` in more places) is to fit better the eHive paradigm. The table name is supposed to be defined in `default_table_name` and the queries should be made against `$self->table_name()` whenever possible. This change is not necessary for the purpose of this PR. I may submit it as a separate PR against master, or just remove it

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisStatsAdaptor.pm
muffato: The query may return `NULL` indeed. I'll see if I handle this case here or where this method is called (in Queen).

muffato:
I am closing this pull request because it should not be merged as it is. As sais in my earlier comments, this needs some more work.

The biggest issue is that I take decisions based on the `runtime_msec` values, which aren't reliable wnough. To make correct decisions, beekeeper should also know about (i) the jobs that are running but haven't yet finished, and (ii) the workers that are pending, and they would all require to be tracked and their statistics kept up-to-date. I also looked at better modelling the job runtimes, and was considering a log-normal distribution. I have pushed my working directory under the tag `attic/smart_scheduler` (just one more commit on top of this PR).

PR #52, opened 2018-09-06
Allow the IF NOT EXISTS option of CREATE DATABASE

muffato:
Found this with @marcoooo. `DROP DATABASE` understands `IF EXISTS`, but `CREATE DATABASE` doesn't understand `IF NOT EXISTS`


muffato:
In the end, rather than manipulating the `CREATE DATABASE` command behind the scene, I have considered it is the responsibility of the caller not to do `IF NOT EXISTS` on PostgreSQL, and I have updated the test-suite accordingly.

PR #53, opened 2018-09-07
 Fixed the Params API in Python

muffato:
This fixes another issue I've found with @marcoooo yesterday. `self.param_required('a')` is supposed to raise an exception if `a` is _None_ (like in Perl). In fact, when the behaviour of the param* methods was rationalised in Perl, the decisions should have been applied to the Python implementation, but I forgot to do so.
Both implementations should now be consistent and in sync with the manual: https://ensembl-hive.readthedocs.io/en/master/creating_runnables/IO_and_errors.html#parameter-handling

### DIFF
--- a/scripts/travis_run_tests.sh
+++ b/scripts/travis_run_tests.sh
@@ -26,9 +26,12 @@ if [ "$COVERALLS" = 'true' ]; then
 else
   prove -r t
 fi
-
 rt=$?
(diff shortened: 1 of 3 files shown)


ens-bwalts:
It looks like this is the case in the Java wrappers as well.

muffato:
OK, right. That will be another pull-request then, but on 2.5.

PR #56, opened 2018-10-26
use information schema

kiwiroy:
## Problem solved

When initialising a pipeline `has_write_access` errors as my user doesn't have `select` privileges on the `mysql.user` table.

    SELECT command denied to user 'kiwiroy'@'hostname.co.nz' for table 'user'

**Edit** The [# TODO](https://github.com/Ensembl/ensembl-hive/blob/c1c364dfd79a1f3375889bf4d11919aaf53db68a/modules/Bio/EnsEMBL/Hive/DBSQL/DBConnection.pm#L386) is attended to in e7f5881.

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/DBSQL/DBConnection.pm
+++ b/modules/Bio/EnsEMBL/Hive/DBSQL/DBConnection.pm
@@ -383,7 +383,7 @@ sub has_write_access {
         my $access_sql =
             q{SELECT COUNT(*)
                 FROM information_schema.user_privileges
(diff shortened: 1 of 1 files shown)


ens-bwalts:
This looks like an improved way to check for privileges, and thanks for adding an appropriate Postgres check. However, it should also check for UPDATE privileges.

kiwiroy:
@ens-bwalts sorry, my bad for translating `SELECT Insert_priv, Update_priv, Delete_priv` incorrectly. Fixed now.

PR #57, opened 2018-10-29
Cache CPAN dependencies for faster builds

kiwiroy:
While the decrease in the time taken to test on travis ([before](https://travis-ci.org/kiwiroy/ensembl-hive/builds/447572596) vs [after](https://travis-ci.org/kiwiroy/ensembl-hive/builds/447600045)) isn't as significant as I hoped there is definitely a benefit that the log files are shorter. The ~11 year old 5.10 still dominates the time taken for the integration suite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,13 @@ addons:
     - ant
     - graphviz
 
+cache:
+  directories:
+    - $HOME/deps
+
 before_install:
+    - cpanm -nq local::lib
+    - eval "$(perl -Mlocal::lib=${HOME}/deps)"
     - git clone --branch v1.6.x --depth 1 https://github.com/bioperl/bioperl-live
     - mysql -u root -h localhost -e 'GRANT ALL PRIVILEGES ON *.* TO "travis"@"%"'
     - mysql -hlocalhost -utravis -e "SET GLOBAL sql_mode = 'TRADITIONAL'"
--- a/scripts/dev/travis_run_tests.sh
+++ b/scripts/dev/travis_run_tests.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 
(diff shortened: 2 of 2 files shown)


muffato:
Thank you @kiwiroy . I didn't know this feature, it looks cool !
Reading https://docs.travis-ci.com/user/caching/#pull-request-builds-and-caches the cache built for this pull-request (#57) will remain specific to this pull-request. It's only if you add another commit to this pull-request that we can see the difference. Then, once we have accepted the pull-request, a new cache will be created for the branch master, which can be used for further builds on that branch.

Also, since all of `$HOME/deps` is cached, I fear that https://github.com/Ensembl/ensembl-hive/blob/version/2.5/.travis.yml#L39 is going to fail unless it's the `-f` option is added too.

muffato:
I've restarted this build https://travis-ci.org/Ensembl/ensembl-hive/jobs/447679572 and it did find and unpacked the cached deps/. I don't understand why `mkdir deps` didn't fail (it needs the `-p` option if the directory already exists) and why `ln` also didn't complain ...
However the `cpanm` commands could all reuse the cache :+1: 

Am I missing something ?

kiwiroy:
@muffato `mkdir deps` is in `pwd` which is something like [`/home/travis/owner/ensembl-hive`](https://travis-ci.org/kiwiroy/ensembl-hive/jobs/447600047#L448) and the main reason I chose `$HOME/deps`, although it required this [addition](https://github.com/Ensembl/ensembl-hive/pull/57/files#diff-b555268d4bf90854a152b5bba27dbd86R18). The caching is as you explain.

muffato:
Ah, got it ! A subtle difference that I missed ... :)
Thanks again @kiwiroy !

muffato:
The same trick also helps speeding up the ensembl-compara builds ;)

kiwiroy:
@muffato awesome.

PR #58, opened 2018-12-10
Protect db_handle methods against "MySQL server has gone away"

muffato:
# Use case

The issue was raised by @MatBarba . In ENSCORESW-2913 he managed to identify that a Python Runnable that just sleeps would raise the following error after 30 min:
```
DBD::mysql::db primary_key failed: MySQL server has gone away at (...)/modules/Bio/EnsEMBL/Hive/DBSQL/BaseAdaptor.pm line 244, <$CHILD_RDR> line 6.
```
30 min is indeed the timeout parameter of the server he was using.

# Description of the changes

The `primary_key` "dbh" method (and a few others) were indeed called on database handles without checking if the handle is still valid, and without any `eval {}` protection, even though such protection is available via the `AUTOLOAD` in `DBSQL::DBConnection`. The fix is to move those calls from the database handle to the database connection (new "best practice").

`prepare` is slightly more tricky to fix since there is already a `prepare` in `DBConnection`. In this case, I have introduced a fake DBConnection method named `protected_prepare` that calls `prepare` on the database handle with the "MySQL server has gone away" protection, but without disconnecting from the database even if `disconnect_when_inactive` is set. Indeed, `prepare` is most often called on SELECT statements, which are immediately followed by `execute` and `fetch`, both of which expect the database connection to still be open

# Comments

I have added a few comments about pitfalls of the current implementation

# Notes

The change in `has_write_access` will probably clash with the changes that have been accepted upstream. I can easily fix it.

### DIFF
--- a/docs/dev/development_guidelines.rst
+++ b/docs/dev/development_guidelines.rst
@@ -57,6 +57,28 @@ Core type. eHive has developed new functionalities in its own
 ``DBSQL::DBConnection``, and has reimplemented ``DBSQL::StatementHandle``
 to allow catching connection errors and automatically reconnecting.
 
+In summary, eHive connections are protected against:
+
+* "MySQL server has gone away" errors through:
+
+  * a reimplementation of ``DBSQL::StatementHandle`` (which allows code to be
+    entirely compatible with the Core API)
+  * allowing to call *dbh* methods on the connection object itself (whilst
+    capturing the error), though this requires the calling code to be updated.
+    In hindsight, it would have been better to implement a new class
+    ``DBSQL::DatabaseHandle`` that would wrap and protect *db_handle* calls the
+    same way ``DBSQL::StatementHandle`` wraps and protects *sth* calls.
+
+    .. note::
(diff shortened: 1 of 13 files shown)


muffato:
Ah, good point though. I didn't think of that. Here is a summary of how the different bits are connected.
- `DBConnection` has `protected_prepare_execute` which runs `prepare`+`execute` and protects primarily against deadlocks and lock timeouts, but also inherits the protection described below
- `prepare` will create a new `StatementHandle` with the lower-level `protected_prepare`
- `CoreDBConnection` (`DBConnection`'s parent class) has `protected_prepare` via `AUTOLOAD`. The latter is used to call _dbh_ methods directly on the DBConnection, catching "MySQL server has gone away errors"
- `StatementHandle` also has an `AUTOLOAD` to pass all its method calls to the underlying _sth_, also catching "MySQL server has gone away errors"

I may add that to the developer documentation ?

muffato:
I'v added some notes in the developer documentation.

PR #61, opened 2019-01-23
Make analysis tags work like logic names

tgrego:
See https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-2558

When runWorker, tweak, etc. try to find analysis, they use `Bio::EnsEMBL::Hive::Utils::Collection->find_all_by_pattern()`.
This ends up defaulting to name.

This pull request adds a fall over search for tags, if no matches exist for name.
Works for searches of a 'tag==word', 'word' and as well as with wildcard 'word%', and is implemented for the collectable that implements tags (i.e. Analysis).

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Analysis.pm
+++ b/modules/Bio/EnsEMBL/Hive/Analysis.pm
@@ -46,7 +46,7 @@ use Bio::EnsEMBL::Hive::GuestProcess;
 
 
 use base ( 'Bio::EnsEMBL::Hive::Storable' );
- 
+
 
 sub unikey {    # override the default from Cacheable parent
     return [ 'logic_name' ];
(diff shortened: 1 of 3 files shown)


muffato:
I wouldn't like to make the default word search (e.g. `analysis[bla]`) include tags. It will be confusing when `bla` would match a different set of analyses if considering analysis names or tags.

In fact, tags are already already searchable, but only exact search is possible, e.g. to find all the analyses with the _tags_ property equal to `trimming`
```tweak_pipeline.pl -url $EHIVE_URL -tweak 'analysis[tags==trimming].module?'```

To me, what is missing is a way of doing regex searches (like `analysis[bla%]` for analysis names) in the Collection against any field. For that, I would suggest to add a new operator in the `Bio::EnsEMBL::Hive::Utils::Collection::_find_all_by_subpattern`: `~`, e.g. `analysis[tags~trim]` or `analysis[module~Bio::EnsEMBL::Production]`.

tgrego:
Thanks for the comments.
I agree with that, it is better to just allow regex search for any field, and keep the word search limited to the field name.
I'm closing this PR and submitting a new one shortly with an attempt at getting that implementation.

PR #62, opened 2019-01-23
Tweak pipeline to json

mira13:
Format tweak-pipeline output with JSON, test included.


muffato:
To be honest, I don't think this makes parsing the output any easier. Trying `tweak_pipeline.pl -url $EHIVE_URL -tweak 'analysis[infernal].analysis_capacity?'`:

_before_
```
Tweak.Request   analysis[infernal].analysis_capacity?
Tweak.Found     1 analyses matching the pattern 'infernal'
Tweak.Show      analysis[infernal].analysis_capacity :: 200
```

_after_
```
{
"Show": "analysis[infernal].analysis_capacity ::200",
"Request": "analysis[infernal].analysis_capacity?",
"Found": "1 analyses matching the pattern 'infernal'"
}
```

Someone still has to parse `analysis[infernal].analysis_capacity ::200`, which is what Thibaut's request was initially about. I mentioned JSON because it allows to structure the output, and can be directly plugged into guiHive so that we can remove some of its custom scripts. I don't know what such APIs typically return, but I would have expected something like
```
{
  "Object": {
    "Type": "Analysis",
    "dbID": 34,
    "name": "infernal"
  },
  "Action": "query",
  "Return": {
    "Field": "analysis_capacity",
    "Value": 200
  }
}
```
where all the information can be directly accessed without any string parsing

mira13:
@muffato  @ens-bwalts  Request is now ready for review, thanks

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -841,16 +941,25 @@ sub apply_tweaks {
                             'submission_cmd_args'   => $new_submission_cmd_args,
                             'worker_cmd_args'       => $new_worker_cmd_args,
                         );
+                        $tweakStructure->{Return}->{OldValue} = '';
                     }
+                    $tweakStructure->{Return}->{Field} = "resource_class[$rc_name].meadow";
+                      $tweakStructure->{Return}->{NewValue} = stringify([$new_submission_cmd_args, $new_worker_cmd_args]);
```
muffato: Indentation

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +829,86 @@ sub apply_tweaks {
...
                                $analysis_stats->is_excluded() . " --> $new_value_str\n";
                            $analysis_stats->is_excluded($new_value);
                            $need_write = 1;
                         }
                     }
                 } elsif($analysis->can($attrib_name)) {
+                    $tweakStructure->{Return}->{Field} = $attrib_name;
                     my $old_value = stringify($analysis->$attrib_name());
-
+                      $tweakStructure->{Return}->{OldValue} = $old_value;
```
muffato: Indentation

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
 sub apply_tweaks {
     my $self    = shift @_;
     my $tweaks  = shift @_;
-
+    my @responce;
+    my $responceStructure;
```
muffato: Initialize this to an empty hash

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
 sub apply_tweaks {
     my $self    = shift @_;
     my $tweaks  = shift @_;
-
+    my @responce;
```
muffato: Should it not be `response` ? (and `responseStructure` below)

modules/Bio/EnsEMBL/Hive/Scripts/InitPipeline.pm
```diff
@@ -62,8 +62,9 @@ sub init_pipeline {
 
     if($tweaks and @$tweaks) {
         print "> Applying tweaks.\n";
-        $pipeline->apply_tweaks( $tweaks );
-        print "\n";
+          my ($need_write, $listRef, $json) = $pipeline->apply_tweaks( $tweaks );
+         print join("\n", @$listRef);
+         print "\n";
```
muffato: Indentation

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
...
+    my $responceStructure;
     my $need_write = 0;
-
+    $responceStructure->{Tweaks} = [];
     foreach my $tweak (@$tweaks) {
-        print "\nTweak.Request\t$tweak\n";
+        push @responce, "Tweak.Request  \t$tweak\n";
```
muffato: Why are you adding a blank space in all those strings ?

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
...
 
         if($tweak=~/^pipeline\.param\[(\w+)\](\?|#|=(.+))$/) {
             my ($param_name, $operator, $new_value_str) = ($1, $2, $3);
-
+            my $tweakStructure;
+            $tweakStructure->{Object}->{Type} = 'Pipeline';
+            $tweakStructure->{Object}->{Id} = '';
```
muffato: I would default to _undef_ to mimic an INT column that can have NULL

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
...
             my $pwp_collection  = $self->collection_of( 'PipelineWideParameters' );
             my $hash_pair       = $pwp_collection->find_one_by('param_name', $param_name);
-
             if($operator eq '?') {
-                print "Tweak.Show    \tpipeline.param[$param_name] ::\t"
-                     . ($hash_pair ? $hash_pair->{'param_value'} : '(missing_value)') . "\n";
+                my $value = $hash_pair ? $hash_pair->{'param_value'} : '(missing_value)';
+                $tweakStructure->{Action} = 'SHOW';
+                $tweakStructure->{Return}->{OldValue} = $value;
+                $tweakStructure->{Return}->{NewValue} = $value;
```
muffato: For SHOW actions, `NewValue` is redundant with `OldValue`. Can we just use `Value` ?

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
...
+                push @responce, "Tweak.Show  \tpipeline.param[$param_name] ::\t"
+                     . ($value) . "\n";
```
muffato: Join to the previous line

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
...
             } elsif($operator eq '#') {
+                $tweakStructure->{Action} = 'DELETE';
+                $tweakStructure->{Return}->{OldValue} = $hash_pair->{'param_value'};
+                $tweakStructure->{Return}->{NewValue} = '';
```
muffato: `''` is not an accurate representation of the new value (neither would _undef_ be). Can be just have `OldValue` for DELETE actions ?

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
...
-
+    $responceStructure->{Tweaks} = [];
     foreach my $tweak (@$tweaks) {
-        print "\nTweak.Request\t$tweak\n";
+        push @responce, "Tweak.Request  \t$tweak\n";
 
         if($tweak=~/^pipeline\.param\[(\w+)\](\?|#|=(.+))$/) {
             my ($param_name, $operator, $new_value_str) = ($1, $2, $3);
-
+            my $tweakStructure;
```
muffato: Since this will be used in all cases of the `if`, can be definition be moved a level up ?
Also, to avoid auto-vivification, can we at least set it to an empty hash ?

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
...
+            push @responce, "Tweak.Found  \t".scalar(@$analyses)." analyses matching the pattern '$analyses_pattern'\n";
 
             my $new_value = destringify( $new_value_str );
 
             foreach my $analysis (@$analyses) {
-
+                my $tweakStructure;
+                $tweakStructure->{Object}->{Type} = "Analysis";
                 my $analysis_name = $analysis->logic_name;
+                $tweakStructure->{Object}->{Id} = $analysis->dbID . " " . $analysis->logic_name;
```
muffato: Same comment as above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
...
 
             foreach my $analysis (@$analyses) {
-                my $analysis_name = $analysis->logic_name;
+                my $tweakStructure;
+                $tweakStructure->{Object}->{Type} = 'Analysis';
 
+                my $analysis_name = $analysis->logic_name;
                 my $old_value = $analysis->parameters;
 
+                $tweakStructure->{Object}->{Id} = $analysis->dbID . " " . $analysis_name;
```
muffato: Since the goal is to have a parsable output, why do you merge fields ? I would suggest to put the analysis' dbID in `Id` and its logic_name in `name`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,203 @@ sub print_diagram {
...
                     );
                 }
             }
-
+          push @{$responceStructure->{Tweaks}}, $tweakStructure;
         } elsif($tweak=~/^pipeline\.(\w+)(\?|=(.+))$/) {
+            my $tweakStructure;
             my ($attrib_name, $operator, $new_value_str) = ($1, $2, $3);
+            $tweakStructure->{Object}->{Type} = 'Pipeline';
+            $tweakStructure->{Object}->{Id} = '';
```
muffato: Same comment as above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -700,58 +752,73 @@ sub apply_tweaks {
...
+            push @responce, "Tweak.Found  \t".scalar(@$analyses)." analyses matching the pattern '$analyses_pattern'\n";
 
             my $new_value = destringify( $new_value_str );
 
             foreach my $analysis (@$analyses) {
 
                 my $analysis_name = $analysis->logic_name;
+                my $tweakStructure;
+                $tweakStructure->{Object}->{Type} = 'Analysis';
+                $tweakStructure->{Object}->{Id} = $analysis->dbID . " " . $analysis_name;
```
muffato: Same comment as above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +829,86 @@ sub apply_tweaks {
...
                 foreach my $rc (@$resource_classes) {
                     my $rc_name = $rc->name;
+                    $tweakStructure->{Object}->{Id} = $rc_name;
+                    $tweakStructure->{Action} = "SHOW";
 
                     if(my $rd = $self->collection_of( 'ResourceDescription' )->find_one_by('resource_class', $rc, 'meadow_type', $meadow_type)) {
                         my ($submission_cmd_args, $worker_cmd_args) = ($rd->submission_cmd_args, $rd->worker_cmd_args);
-                        print "Tweak.Show    \tresource_class[$rc_name].$meadow_type ::\t".stringify([$submission_cmd_args, $worker_cmd_args])."\n";
+                        push @responce, "Tweak.Show  \tresource_class[$rc_name].$meadow_type ::\t".stringify([$submission_cmd_args, $worker_cmd_args])."\n";
+                          $tweakStructure->{Return}->{OldValue} = stringify([$submission_cmd_args, $worker_cmd_args]);
```
muffato: Indentation

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: We need every iteration be sure that $tweakStructure in new (doesn't contain any previous result), so I create it every time locally;

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: To make output consistent, all output lines to have the same space after action string

t/03.scripts/tweak_pipeline_output.t
```diff
@@ -0,0 +1,122 @@
...
+    push @tweak_requests,  ["-tweak" => "resource_class[urgent].LSF?"];
+
+    foreach my $request (@tweak_requests) {
+      my $stdout = capture_stdout {
+          tweak_pipeline($pipeline_url, $request);
+      };
+      is_valid_json $stdout;
+      my $stdoutJson = decode_json($stdout);
+      use Data::Dumper;
+      ok(scalar @{$stdoutJson->{Tweaks}} > 0, "Tweaks responce recieved for " . join (' ', @{$request}) . Dumper($stdoutJson->{Tweaks}));
```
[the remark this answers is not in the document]
mira13: I've done it consistent with all other tests here, they all are using test msg, not diag

mira13:
@muffato all comment are resolved (or commented back), please check

scripts/tweak_pipeline.pl
```diff
@@ -65,9 +66,13 @@ sub main {
...
     }
-
     if(@$tweaks) {
-        my $need_write = $pipeline->apply_tweaks( $tweaks );
+        my ($need_write, $listRef, $responceStructure) = $pipeline->apply_tweaks( $tweaks );
+
+        $responceStructure->{URL} = $self->{'url'};
+        my $json = JSON->new->allow_nonref;
+
+        print $self->{'json'} ? $json->encode($responceStructure) : join(', ', @$listRef);
```
ens-bwalts: The ', ' is not needed here, since the plain text output has '\n' in it.

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
ens-bwalts: I agree with @muffato here.

```
$tweakStructure->{Object}->{Id} = $analysis->dbID;
$tweakStructure->{Object}->{Name} = $analysis_name;
# similar for resource class
# for pipeline, can provide undef for both.
```

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,206 @@ sub print_diagram {
...
+                $tweakStructure->{Object}->{Type} = 'Analysis';
 
+                my $analysis_name = $analysis->logic_name;
                 my $old_value = $analysis->parameters;
 
+                $tweakStructure->{Object}->{Id} = $analysis->dbID;
+                $tweakStructure->{Object}->{Name} = $analysis_name;
+                $tweakStructure->{Return}->{Field} = $param_name;
                 my $param_hash  = destringify( $old_value );
+                $tweakStructure->{Return}->{OldValue} =  exists($param_hash->{ $param_name }) ? stringify($param_hash->{ $param_name }) : '';
```
muffato: _undef_, not an empty string

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -700,58 +755,74 @@ sub apply_tweaks {
...
             foreach my $analysis (@$analyses) {
 
                 my $analysis_name = $analysis->logic_name;
+                my $tweakStructure;
+                $tweakStructure->{Object}->{Type} = 'Analysis';
+                $tweakStructure->{Object}->{Id} = $analysis->dbID;
+                $tweakStructure->{Object}->{Name} = $analysis_name;
 
                 if( $attrib_name eq 'resource_class' ) {
+                  $tweakStructure->{Return}->{OldValue} = $analysis->resource_class ? $analysis->resource_class->name : '';
```
muffato: _undef_, not an empty string

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,206 @@ sub print_diagram {
...
-                print "Tweak.Show    \tpipeline.param[$param_name] ::\t"
-                     . ($hash_pair ? $hash_pair->{'param_value'} : '(missing_value)') . "\n";
+                my $value = $hash_pair ? $hash_pair->{'param_value'} : '(missing_value)';
+                $tweakStructure->{Action} = 'SHOW';
+                $tweakStructure->{Return}->{OldValue} = $value;
+                $tweakStructure->{Return}->{NewValue} = $value;
+                push @response, "Tweak.Show  \tpipeline.param[$param_name] ::\t" . ($value) . "\n";
             } elsif($operator eq '#') {
+                $tweakStructure->{Action} = 'DELETE';
+                $tweakStructure->{Return}->{OldValue} = $hash_pair->{'param_value'};
```
muffato: `$hash_pair` can be undefined

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,206 @@ sub print_diagram {
...
+            $tweakStructure->{Return}->{Field} = $param_name;
             my $pwp_collection  = $self->collection_of( 'PipelineWideParameters' );
             my $hash_pair       = $pwp_collection->find_one_by('param_name', $param_name);
-
             if($operator eq '?') {
-                print "Tweak.Show    \tpipeline.param[$param_name] ::\t"
-                     . ($hash_pair ? $hash_pair->{'param_value'} : '(missing_value)') . "\n";
+                my $value = $hash_pair ? $hash_pair->{'param_value'} : '(missing_value)';
+                $tweakStructure->{Action} = 'SHOW';
+                $tweakStructure->{Return}->{OldValue} = $value;
```
muffato: I don't think we should show `(missing_value)` in the JSON output. _undef_  instead ?

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -841,16 +947,25 @@ sub apply_tweaks {
...
                             'worker_cmd_args'       => $new_worker_cmd_args,
                         );
+                        $tweakStructure->{Return}->{OldValue} = undef;
                     }
+                    $tweakStructure->{Return}->{Field} = "resource_class[$rc_name].meadow";
+                    $tweakStructure->{Return}->{NewValue} = stringify([$new_submission_cmd_args, $new_worker_cmd_args]);
                     $need_write = 1;
                 }
             }
+            push @{$responseStructure->{Tweaks}}, $tweakStructure;
```
muffato: `$tweakStructure` needs to be pushed at every iteration of the for loop on `@resource_classes`, otherwise we only get the last tweak

t/03.scripts/tweak_pipeline_output.t
```diff
@@ -0,0 +1,114 @@
...
+  my $stdoutJson = decode_json($stdout);
+  use Data::Dumper;
+  ok(scalar @{$stdoutJson->{Tweaks}} > 0, "Tweaks responce recieved for " . join (' ', @{$request}) . Dumper($stdoutJson->{Tweaks}));
+  foreach my $tweakJson (@{$stdoutJson->{Tweaks}}) {
+    ok($tweakJson->{Object}->{Type} eq "Pipeline" ||
+      $tweakJson->{Object}->{Type} eq "Analysis" ||
+      $tweakJson->{Object}->{Type} eq "Resource class",
+      'Object type is correct for ' . join (' ', @{$request}));
+
+    ok( $tweakJson->{Error} ||
```
muffato: So basically if there is an error, nothing in the JSON can be relied upon apart from Object>Type. I think we can guarantee more than that. I would say that the output structure should always have `Action` and `Object`, and then either `Return` or `Error`. This means going back to HivePipeline.pm and adding the missing fields whenever there is an error.

mira13: If we put "delete" action - and it's not expected, we can't put it as action: "delete" error: ..., that would mean delete error, but the error is about wrong parameter. The same about object - it would mean the error is about attempt on the object - it can't be deleted, but the error is just about this is forbidden parameters combination

mira13: There are some places where we can tell that the action was - "DELETE" and in that places we now fill the action, but there are also places (Failed to parcel tweak at line 963) where we can't fill the action - that's why we can't guarantee action to be there.

muffato: OK, so the JSON structure and the test should follow this reasoning:
1. Can we parse the tweak ?
   * NO: output is a JSON with _only_ `error`
   * YES: we populate the JSON with `Object` and `Action`. Then, 2) can we execute the action ?
      * NO: add `error`
      * YES: add `Return`

scripts/tweak_pipeline.pl
```diff
@@ -65,9 +66,13 @@ sub main {
     } else {
         die "\nERROR: Connection parameters (url or reg_conf+reg_alias) need to be specified\n";
     }
-
     if(@$tweaks) {
-        my $need_write = $pipeline->apply_tweaks( $tweaks );
+        my ($need_write, $listRef, $responceStructure) = $pipeline->apply_tweaks( $tweaks );
```
muffato: Same comment here as in Scripts/InitPipeline.pm: snake case is the norm. For consistency use the same variable names in both files. However, `$json` would clash with the instance of the JSON codec below, so rename `$json` in one of the two files

modules/Bio/EnsEMBL/Hive/Scripts/InitPipeline.pm
```diff
@@ -62,7 +62,8 @@ sub init_pipeline {
 
     if($tweaks and @$tweaks) {
         print "> Applying tweaks.\n";
-        $pipeline->apply_tweaks( $tweaks );
+        my ($need_write, $msg_listRef, $json) = $pipeline->apply_tweaks( $tweaks );
```
muffato: `$msg_listRef` is funny because it mixes snake case and camel case. The whole file is in snake case, so please use `$msg_list_ref`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +833,88 @@ sub apply_tweaks {
...
-
+            $tweakStructure->{Object}->{Type} = "Resource class";
             my $resource_classes = $self->collection_of( 'ResourceClass' )->find_all_by_pattern( $rc_pattern );
-            print "Tweak.Found   \t".scalar(@$resource_classes)." resource_classes matching the pattern '$rc_pattern'\n";
+            push @response, "Tweak.Found  \t".scalar(@$resource_classes)." resource_classes matching the pattern '$rc_pattern'\n";
 
             if($operator eq '?') {
                 foreach my $rc (@$resource_classes) {
                     my $rc_name = $rc->name;
+                    $tweakStructure->{Object}->{Id} = $rc->dbID;
```
muffato: `$tweakStructure` needs to be reset every loop otherwise the `Tweaks` list will contain references to the same object

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -822,17 +924,22 @@ sub apply_tweaks {
 
                 foreach my $rc (@$resource_classes) {
                     my $rc_name = $rc->name;
+                    $tweakStructure->{Object}->{Id} = $rc->dbID;
```
muffato: `$tweakStructure` needs to be reset every loop otherwise the `Tweaks` list will contain references to the same object

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,206 @@ sub print_diagram {
...
             my $hash_pair       = $pwp_collection->find_one_by('param_name', $param_name);
-
             if($operator eq '?') {
-                print "Tweak.Show    \tpipeline.param[$param_name] ::\t"
-                     . ($hash_pair ? $hash_pair->{'param_value'} : '(missing_value)') . "\n";
+                my $value = $hash_pair ? $hash_pair->{'param_value'} : undef;
+                $tweakStructure->{Action} = 'SHOW';
+                $tweakStructure->{Return}->{OldValue} = $value;
+                $tweakStructure->{Return}->{NewValue} = $value;
+                push @response, "Tweak.Show  \tpipeline.param[$param_name] ::\t" . $value ? $value : '--> (missing value)' . "\n";
```
muffato: * Remove `-->`
* `$value` can be an empty string

so do `$value // '(missing value)'`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -700,58 +755,74 @@ sub apply_tweaks {
...
                         if(my $old_value = $analysis->resource_class) {
-                            print "Tweak.Show    \tanalysis[$analysis_name].resource_class ::\t".$old_value->name."\n";
+                            push @response, "Tweak.Show  \tanalysis[$analysis_name].resource_class ::\t".$old_value->name."\n";
                         } else {
-                            print "Tweak.Show    \tanalysis[$analysis_name].resource_class ::\t(missing value)\n";
+                            push @response, "Tweak.Show  \tanalysis[$analysis_name].resource_class ::\t(missing value)\n";
                         }
                     } elsif($operator eq '#') {
-                        print "Tweak.Error   \tDeleting of ResourceClasses is not supported\n";
+                        my $errorMsg = 'Tweak.Error  \tDeleting of ResourceClasses is not supported\n';
```
muffato: Need double-quotes because of `\t`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: I have just left the same msg as it was before, not to confuse those who already use it

mira13:
@muffato now ACTION and OBJECT is guaranteed, and error msgs for JSON simplified

scripts/tweak_pipeline.pl
```diff
@@ -65,9 +66,13 @@ sub main {
...
     }
-
     if(@$tweaks) {
-        my $need_write = $pipeline->apply_tweaks( $tweaks );
+        my ($need_write, $msg_list_ref, $responceStructure) = $pipeline->apply_tweaks( $tweaks );
+
+        $responceStructure->{URL} = $self->{'url'};
+        my $json = JSON->new->allow_nonref;
+
+        print $self->{'json'} ? $json->encode($responceStructure) : join('', @$$msg_list_ref);
```
muffato: Only 1 `$` should enough in front of `msg_list_ref`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,210 @@ sub print_diagram {
...
 
             foreach my $analysis (@$analyses) {
-                my $analysis_name = $analysis->logic_name;
+                my $tweakStructure;
+                $tweakStructure->{Object}->{Type} = 'Analysis';
 
+                my $analysis_name = $analysis->logic_name;
                 my $old_value = $analysis->parameters;
 
+                $tweakStructure->{Object}->{Id} = $analysis->dbID;
```
muffato: Can you do `$analysis->dbID + 0` to force it being considered as a numeric ?

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,210 @@ sub print_diagram {
...
+            push @response, "Tweak.Found  \t".scalar(@$analyses)." analyses matching the pattern '$analyses_pattern'\n";
 
             my $new_value = destringify( $new_value_str );
 
             foreach my $analysis (@$analyses) {
-
+                my $tweakStructure;
+                $tweakStructure->{Object}->{Type} = "Analysis";
                 my $analysis_name = $analysis->logic_name;
+                $tweakStructure->{Object}->{Id} = $analysis->dbID;
```
muffato: Add `+ 0` like above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -700,58 +759,74 @@ sub apply_tweaks {
...
+            push @response, "Tweak.Found  \t".scalar(@$analyses)." analyses matching the pattern '$analyses_pattern'\n";
 
             my $new_value = destringify( $new_value_str );
 
             foreach my $analysis (@$analyses) {
 
                 my $analysis_name = $analysis->logic_name;
+                my $tweakStructure;
+                $tweakStructure->{Object}->{Type} = 'Analysis';
+                $tweakStructure->{Object}->{Id} = $analysis->dbID;
```
muffato: Add `+ 0` like above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +837,93 @@ sub apply_tweaks {
...
             my $resource_classes = $self->collection_of( 'ResourceClass' )->find_all_by_pattern( $rc_pattern );
-            print "Tweak.Found   \t".scalar(@$resource_classes)." resource_classes matching the pattern '$rc_pattern'\n";
+            push @response, "Tweak.Found  \t".scalar(@$resource_classes)." resource_classes matching the pattern '$rc_pattern'\n";
 
             if($operator eq '?') {
                 foreach my $rc (@$resource_classes) {
+                    my $tweakStructure;
+                    $tweakStructure->{Object}->{Type} = "Resource class";
                     my $rc_name = $rc->name;
+                    $tweakStructure->{Object}->{Id} = $rc->dbID;
```
muffato: Add `+ 0` like above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -821,18 +932,25 @@ sub apply_tweaks {
                 my ($new_submission_cmd_args, $new_worker_cmd_args) = (ref($new_value) eq 'ARRAY') ? @$new_value : ($new_value, '');
 
                 foreach my $rc (@$resource_classes) {
+                    my $tweakStructure;
+                    $tweakStructure->{Object}->{Type} = "Resource class";
                     my $rc_name = $rc->name;
+                    $tweakStructure->{Object}->{Id} = $rc->dbID;
```
muffato: Add `+ 0` like above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,210 @@ sub print_diagram {
...
                 }
 
             } else {
-                print "Tweak.Error   \tCould not find the pipeline-wide '$attrib_name' method\n";
+                if($operator eq '?') {
+                    $tweakStructure->{Action} = 'SHOW';
+                } else {
+                    $tweakStructure->{Action} = 'SET';
+                }
+                $tweakStructure->{Error} = 'Field not recognized';
```
muffato: Since the same messages are going to be used for several errors, define the four messages outside of the function and use the variables here. It will be easier to maintain and will guarantee consistency.

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -700,58 +759,74 @@ sub apply_tweaks {
...
-                            print "Tweak.Show    \tanalysis[$analysis_name].resource_class ::\t".$old_value->name."\n";
+                            push @response, "Tweak.Show  \tanalysis[$analysis_name].resource_class ::\t".$old_value->name."\n";
                         } else {
-                            print "Tweak.Show    \tanalysis[$analysis_name].resource_class ::\t(missing value)\n";
+                            push @response, "Tweak.Show  \tanalysis[$analysis_name].resource_class ::\t(missing value)\n";
                         }
                     } elsif($operator eq '#') {
-                        print "Tweak.Error   \tDeleting of ResourceClasses is not supported\n";
+                        $tweakStructure->{Action} = "DELETE";
+                        $tweakStructure->{Error} = "Action is not supported";
```
muffato: Define outside of `apply_tweaks`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +837,93 @@ sub apply_tweaks {
...
+                    $tweakStructure->{Return}->{OldValue} = $analysis_stats->is_excluded();
                     if($operator eq '?') {
-                        print "Tweak.Show    \tanalysis[$analysis_name].is_excluded ::\t".$analysis_stats->is_excluded()."\n";
+                        $tweakStructure->{Action} = "SHOW";
+                        $tweakStructure->{Return}->{NewValue} = $tweakStructure->{Return}->{OldValue};
+                        push @response, "Tweak.Show  \tanalysis[$analysis_name].is_excluded ::\t".$analysis_stats->is_excluded()."\n";
                     } elsif($operator eq '#') {
-                        print "Tweak.Error   \tDeleting of excluded status is not supported\n";
+                        $tweakStructure->{Action} = "DELETE";
+                        $tweakStructure->{Error} = "Action is not supported";
```
muffato: Define outside of `apply_tweaks`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +837,93 @@ sub apply_tweaks {
...
+                        push @response, "Tweak.Error  \tDeleting of excluded status is not supported\n";
                     } else {
+                        $tweakStructure->{Action} = "SET";
+                        $tweakStructure->{Return}->{NewValue} = $new_value_str;
                         if(!($new_value =~ /^[01]$/)) {
-                            print "Tweak.Error    \tis_excluded can only be 0 (no) or 1 (yes)\n";
+                            $tweakStructure->{Error} = "Invalid value";
```
muffato: Define outside of `apply_tweaks`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +837,93 @@ sub apply_tweaks {
...
+                    $tweakStructure->{Return}->{OldValue} = $old_value;
                     if($operator eq '?') {
-                        print "Tweak.Show    \tanalysis[$analysis_name].$attrib_name ::\t$old_value\n";
+                        $tweakStructure->{Action} = "SHOW";
+                        $tweakStructure->{Return}->{NewValue} = $tweakStructure->{Return}->{OldValue};
+                        push @response, "Tweak.Show  \tanalysis[$analysis_name].$attrib_name ::\t$old_value\n";
                     } elsif($operator eq '#') {
-                        print "Tweak.Error   \tDeleting of Analysis attributes is not supported\n";
+                        $tweakStructure->{Action} = "DELETE";
+                        $tweakStructure->{Error} = "Action is not supported";
```
muffato: Define outside of `apply_tweaks`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +837,93 @@ sub apply_tweaks {
...
                 } else {
-                    print "Tweak.Error   \tAnalysis does not support '$attrib_name' attribute\n";
+                    if($operator eq '?') {
+                        $tweakStructure->{Action} = "SHOW";
+                    } elsif($operator eq '#') {
+                          $tweakStructure->{Action} = "DELETE";
+                    } else {
+                        $tweakStructure->{Action} = "SET";
+                    }
+                    $tweakStructure->{Error} = 'Field not recognized';
```
muffato: Define outside of `apply_tweaks`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -841,16 +959,26 @@ sub apply_tweaks {
...
+                    push @{$responseStructure->{Tweaks}}, $tweakStructure;
                 }
             }
 
+
         } else {
-            print "Tweak.Error   \tFailed to parse the tweak\n";
+            my $tweakStructure;
+            my $errorMsg =
+            $tweakStructure->{Error} = "Tweak cannot be parsed";
```
muffato: Define outside of `apply_tweaks`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -841,16 +959,26 @@ sub apply_tweaks {
...
                     $need_write = 1;
+                    push @{$responseStructure->{Tweaks}}, $tweakStructure;
                 }
             }
 
+
         } else {
-            print "Tweak.Error   \tFailed to parse the tweak\n";
+            my $tweakStructure;
+            my $errorMsg =
```
muffato: Variable not used

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
muffato: _Return > Field_ is missing in this case

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +544,210 @@ sub print_diagram {
...
                     $need_write = 1;
                 }
 
             } else {
-                print "Tweak.Error   \tCould not find the pipeline-wide '$attrib_name' method\n";
+                if($operator eq '?') {
+                    $tweakStructure->{Action} = 'SHOW';
+                } else {
+                    $tweakStructure->{Action} = 'SET';
+                }
```
muffato: There is a similar `if` block below. What about defining outside of the function a hash that maps the operator to the action name ? Then you can use this hash to populate _Action_ in every tweak, not just these two `if` blocks, which would guarantee that _Action_ obeys a controlled vocabulary

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +837,93 @@ sub apply_tweaks {
...
                     }
                 } else {
-                    print "Tweak.Error   \tAnalysis does not support '$attrib_name' attribute\n";
+                    if($operator eq '?') {
+                        $tweakStructure->{Action} = "SHOW";
+                    } elsif($operator eq '#') {
+                          $tweakStructure->{Action} = "DELETE";
+                    } else {
+                        $tweakStructure->{Action} = "SET";
+                    }
```
muffato: Define the mapping outside of the function

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +837,93 @@ sub apply_tweaks {
...
                         my ($submission_cmd_args, $worker_cmd_args) = ($rd->submission_cmd_args, $rd->worker_cmd_args);
-                        print "Tweak.Show    \tresource_class[$rc_name].$meadow_type ::\t".stringify([$submission_cmd_args, $worker_cmd_args])."\n";
+                        push @response, "Tweak.Show  \tresource_class[$rc_name].$meadow_type ::\t".stringify([$submission_cmd_args, $worker_cmd_args])."\n";
+                        $tweakStructure->{Return}->{OldValue} = stringify([$submission_cmd_args, $worker_cmd_args]);
                     } else {
-                        print "Tweak.Show    \tresource_class[$rc_name].$meadow_type ::\t(missing values)\n";
+                        push @response, "Tweak.Show  \tresource_class[$rc_name].$meadow_type ::\t(missing values)\n";
+                        $tweakStructure->{Return}->{OldValue} = undef;
                     }
+                    $tweakStructure->{Return}->{Field} = "resource_class[$rc_name].$meadow_type";
```
muffato: In this case, _Type_ is "Resource class" and _Name is `$rc_name`. I think it's redundant to add them in _Field_ and would set the latter to just `$meadow_type`. This would be consistent with the way analyses are modelled, e.g. `resource_class[urgent].LSF` vs `analysis[1].is_excluded`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -821,18 +932,25 @@ sub apply_tweaks {
...
                                 .stringify([$submission_cmd_args, $worker_cmd_args])." --> "
                                 .stringify([$new_submission_cmd_args, $new_worker_cmd_args])."\n";
 
                         $rd->submission_cmd_args(   $new_submission_cmd_args );
                         $rd->worker_cmd_args(       $new_worker_cmd_args     );
+                        $tweakStructure->{Return}->{OldValue} = stringify([$submission_cmd_args, $worker_cmd_args]);
+
                     } else {
-                        print "Tweak.Adding  \tresource_class[$rc_name].meadow :: (missing values) --> "
+                        push @response, "Tweak.Adding  \tresource_class[$rc_name].meadow :: (missing values) --> "
```
muffato: There was a bug: should be `$meadow_type`, not `meadow`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -821,18 +932,25 @@ sub apply_tweaks {
...
+                    $tweakStructure->{Object}->{Type} = "Resource class";
                     my $rc_name = $rc->name;
+                    $tweakStructure->{Object}->{Id} = $rc->dbID;
+                    $tweakStructure->{Object}->{Name} = $rc_name;
+                    $tweakStructure->{Action} = "SET";
 
                     if(my $rd = $self->collection_of( 'ResourceDescription' )->find_one_by('resource_class', $rc, 'meadow_type', $meadow_type)) {
                         my ($submission_cmd_args, $worker_cmd_args) = ($rd->submission_cmd_args, $rd->worker_cmd_args);
-                        print "Tweak.Changing\tresource_class[$rc_name].meadow :: "
+                        push @response, "Tweak.Changing  \tresource_class[$rc_name].meadow :: "
```
muffato: There was a bug: should be `$meadow_type`, not `meadow`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -841,16 +959,26 @@ sub apply_tweaks {
                             'submission_cmd_args'   => $new_submission_cmd_args,
                             'worker_cmd_args'       => $new_worker_cmd_args,
                         );
+                        $tweakStructure->{Return}->{OldValue} = undef;
                     }
+                    $tweakStructure->{Return}->{Field} = "resource_class[$rc_name].meadow";
```
muffato: Set the field to `$meadow_type`

scripts/tweak_pipeline.pl
```diff
@@ -65,9 +66,13 @@ sub main {
     } else {
         die "\nERROR: Connection parameters (url or reg_conf+reg_alias) need to be specified\n";
     }
-
     if(@$tweaks) {
-        my $need_write = $pipeline->apply_tweaks( $tweaks );
+        my ($need_write, $msg_list_ref, $responceStructure) = $pipeline->apply_tweaks( $tweaks );
```
muffato: Rename `$responceStructure`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +557,197 @@ sub print_diagram {
...
-
+                $tweakStructure->{Object}->{Id} = $analysis->dbID + 0;
+                $tweakStructure->{Object}->{Name} = $analysis->logic_name;
                 if( $attrib_name eq 'wait_for' ) {
-
+                    $tweakStructure->{Return}->{Field} = 'wait_for';
                     my $cr_collection   = $self->collection_of( 'AnalysisCtrlRule' );
                     my $acr_collection  = $analysis->control_rules_collection;
-
+                    $tweakStructure->{Return}->{OldValue} = join(', ', map { $_->condition_analysis_url } @$acr_collection );
```
muffato: Since `Bio::EnsEMBL::Hive::Utils::PCL::parse_wait_for` (mostly) expects an array-ref, `OldValue` should be shown in the same format:
```perl
$tweakStructure->{Return}->{OldValue} = [map { $_->condition_analysis_url } @$acr_collection]; 
```

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +557,197 @@ sub print_diagram {
...
             my $new_value = destringify( $new_value_str );
 
             foreach my $analysis (@$analyses) {
-
+                my $tweakStructure;
+                $tweakStructure->{Object}->{Type} = "Analysis";
                 my $analysis_name = $analysis->logic_name;
-
+                $tweakStructure->{Object}->{Id} = $analysis->dbID + 0;
+                $tweakStructure->{Object}->{Name} = $analysis->logic_name;
```
muffato: Set the Action here rather than in the `if` blocks below

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +557,197 @@ sub print_diagram {
...
                 if( $attrib_name eq 'wait_for' ) {
-
+                    $tweakStructure->{Return}->{Field} = 'wait_for';
```
muffato: _Field_ is simply `$attrib_name`, which can be set outside of the `if` blocks

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +557,197 @@ sub print_diagram {
...
                             $need_write = 1;
 
-                            print "Tweak.Deleting\t".$c_rule->toString." --> (missing value)\n";
+                            push @response, "Tweak.Deleting\t".$c_rule->toString." --> (missing value)\n";
                         }
                     }
 
                     if($operator eq '=' or $operator eq '+=') {     # create new rules
+                        $tweakStructure->{Action} = "SET";
+                        $tweakStructure->{Return}->{NewValue} = $new_value;
```
muffato: When the action is `+=`, _Action_ will be `SET` and _NewValue_ the stuff that is **added**. That doesn't feel right. _NewValue_ should be the new, complete, value of the field, so should be set after calling `parse_wait_for` in the same way as `OldValue`. I wonder whether _Action_ should be something like `ADD` or `EXTEND` (must be consistent across all tweak syntaxes)

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +557,197 @@ sub print_diagram {
...
                 } elsif( $attrib_name eq 'flow_into' ) {
-
+                    $tweakStructure->{Return}->{Field} = 'flow_into';
+                    $tweakStructure->{Return}->{OldValue} = $analysis->logic_name;
                     if($operator eq '?') {
                         # FIXME: should not recurse
-                        $analysis->print_diagram_node($self, '', {});
+                        $tweakStructure->{Action} = 'SHOW';
+                        $tweakStructure->{Return}->{NewValue} = $tweakStructure->{Return}->{OldValue};
+                        #$analysis->print_diagram_node($self, '', {});
```
muffato: `print_diagram_node` is not great because it recurses, but that's the only thing we have.

* Either we keep it, but (1) the FIXME comment must be moved closer and (2) the call must be wrapped something that captures STDOUT in order to add it to `@response`
* Or we admit we cannot print dataflow rules even in text mode, and we push a warning / explanation to `@response`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +557,197 @@ sub print_diagram {
...
+                        $tweakStructure->{Action} = "SET";
+                        $tweakStructure->{Return}->{NewValue} = $new_value;
                         Bio::EnsEMBL::Hive::Utils::PCL::parse_wait_for($self, $analysis, $new_value);
                         $need_write = 1;
                     }
 
                 } elsif( $attrib_name eq 'flow_into' ) {
-
+                    $tweakStructure->{Return}->{Field} = 'flow_into';
+                    $tweakStructure->{Return}->{OldValue} = $analysis->logic_name;
```
muffato: `$analysis->logic_name` is **not** the old value

muffato: I don't think we can populate _OldValue_ here

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +557,197 @@ sub print_diagram {
...
                     if($operator eq '?') {
                         # FIXME: should not recurse
-                        $analysis->print_diagram_node($self, '', {});
+                        $tweakStructure->{Action} = 'SHOW';
+                        $tweakStructure->{Return}->{NewValue} = $tweakStructure->{Return}->{OldValue};
```
muffato: Cannot populate _NewValue_

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +557,197 @@ sub print_diagram {
...
+                        #$analysis->print_diagram_node($self, '', {});
                     }
 
                     if($operator eq '#' or $operator eq '=') {     # delete the existing rules
+                        $tweakStructure->{Action} = 'DELETE';
+                        $tweakStructure->{Return}->{NewValue} = undef;
```
muffato: Since there wouldn't be _OldValue_ at all maybe we can skip setting _NewValue_ here

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -700,58 +759,71 @@ sub apply_tweaks {
...
                                 $need_write = 1;
 
-                                print "Tweak.Deleting\t".$df_rule->toString." --> (missing value)\n";
+                                push @response, "Tweak.Deleting\t".$df_rule->toString." --> (missing value)\n";
                             }
                         }
                     }
 
                     if($operator eq '=' or $operator eq '+=') {     # create new rules
+                        $tweakStructure->{Action} = 'SET';
```
muffato: Consider what action to use for `+=`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -700,58 +759,71 @@ sub apply_tweaks {
...
+                        $tweakStructure->{Return}->{NewValue} = $new_value;
```
muffato: `$new_value` is the only thing we have but since we havne't defined the format for dataflow rules, should it be skipped ?

t/03.scripts/tweak_pipeline_output.t
```diff
@@ -0,0 +1,113 @@
...
+        ok( $tweakJson->{Action} eq "SET" ||
+            $tweakJson->{Action} eq "SHOW" ||
+            $tweakJson->{Action} eq "DELETE",
+            'Action field is correct for ' . join (' ', @{$request}));
+
+        ok(exists ($tweakJson->{Object}->{Id}), 'Id field exists for' . join (' ', @{$request}));
+        ok(exists ($tweakJson->{Object}->{Name}), 'Name field exists for' . join (' ', @{$request}));
+        ok(exists ($tweakJson->{Return}->{OldValue}) || $tweakJson->{Error}, 'Old value exists for ' . join (' ', @{$request}));
+        ok(exists($tweakJson->{Return}->{NewValue}) || $tweakJson->{Error}, 'New value exists for ' . join (' ', @{$request}));
+        ok($tweakJson->{Return}->{Field} || $tweakJson->{Error}, 'Field exists for ' . join (' ', @{$request}));
```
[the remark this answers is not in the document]
mira13: No such cases, I will fix the test

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: Here it has unusual actions, so doesn't make sense to put it above, anyway it needs 

> if

:

>  if($operator eq '#' or $operator eq '=') {    
>                         $tweakStructure->{Action} = 'DELETE';
> 
>  if($operator eq '=' or $operator eq '+=') {     # create new rules
>                         $tweakStructure->{Action} = 'SET';

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: The same with actions above - they have unusual structure

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: There is very poor situation with output at all, in this particular file we have 4 type for output:
1. Direct print
2. Push to warnings with print at the end
3. Push to response that is returned
4. Making a hash, returned also (for further JSON)

That's why I have created (and it's pull request) formatter. 
in Formatter it can be added JSON, error, warning, info or custom output function (as here print_diagram_node)  - and printed in needed format in necessary time.

So in scope of this branch - we resolve problem - to get all print out of HivePipeline, that's why we can't keep here print_diagram_node.

In order to refactor it - we will apply formatter here eventually, and it will allow to keep print_diagram_node

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: We aimed to have the same son structure for all responses so there should be serious reasons to break it.

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: Should set it to guarantee fields

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +558,187 @@ sub print_diagram {
...
                             $cr_collection->forget_and_mark_for_deletion( $c_rule );
                             $need_write = 1;
 
-                            print "Tweak.Deleting\t".$c_rule->toString." --> (missing value)\n";
+                            push @response, "Tweak.Deleting\t".$c_rule->toString." --> (missing value)\n";
                         }
                     }
 
                     if($operator eq '=' or $operator eq '+=') {     # create new rules
+                        $tweakStructure->{Return}->{NewValue} = $new_value;
```
muffato: I think I've already reported this. When the action is `+=`, `$new_value` is the stuff that is _added_. I would expect `NewValue` to contains the complete final value, i.e. what was already there + what's been added

This is what we currently get:
```
$ tweak_pipeline.pl -url $(mysql-ens-compara-prod-6-ensadmin details url muffato_test_plants_goc_96) -tweak 'analysis[2].wait_for+=aaa' -json | json_pp
```
```json
{
   "URL" : "'mysql://ensadmin:${_EHIVE_HIDDEN_PASS}@mysql-ens-compara-prod-6.ebi.ac.uk:4616/muffato_test_plants_goc_96'",
   "Tweaks" : [
      {
         "Object" : {
            "Name" : "backbone_fire_clustering",
            "Id" : 2,
            "Type" : "Analysis"
         },
         "Action" : "SET",
         "Return" : {
            "Field" : "wait_for",
            "NewValue" : "aaa",
            "OldValue" : [
               "backbone_fire_clustering",
               "backbone_fire_db_prepare"
            ]
         }
      }
   ]
}
```

t/03.scripts/tweak_pipeline_output.t
```diff
@@ -0,0 +1,113 @@
...
+            'Object type is correct for ' . join (' ', @{$request}));
+
+        ok( $tweakJson->{Action} eq "SET" ||
+            $tweakJson->{Action} eq "SHOW" ||
+            $tweakJson->{Action} eq "DELETE",
+            'Action field is correct for ' . join (' ', @{$request}));
+
+        ok(exists ($tweakJson->{Object}->{Id}), 'Id field exists for' . join (' ', @{$request}));
+        ok(exists ($tweakJson->{Object}->{Name}), 'Name field exists for' . join (' ', @{$request}));
+        ok(exists ($tweakJson->{Return}->{OldValue}) || $tweakJson->{Warning} || $tweakJson->{Error}, 'Old value exists for ' . join (' ', @{$request}));
```
muffato: Can some of these `||` be made exclusive ?

muffato: Can you change the mark in `@tweak_requests` which ones are expected to fail and test them appropriately here ? Otherwise, all tweaks could be returning an error and the unit-test would still be happy

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -186,6 +186,20 @@ sub new {       # construct an attached or a detached Pipeline object
 #       warn "Created a standalone pipeline";
     }
 
+    $self->{ERROR_MSG} = {
```
muffato: Rename this to `TWEAK_ERROR_MSG`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -186,6 +186,20 @@ sub new {       # construct an attached or a detached Pipeline object
...
+        PARSE_ERROR => "Tweak cannot be parsed",
+        ACTION_ERROR => "Action is not supported",
+        FIELD_ERROR => "Field not recognized",
+        VALUE_ERROR => "Invalid value"
+    };
+
+    $self->{ACTION} = {
```
muffato: Rename this to `TWEAK_ACTION`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +558,187 @@ sub print_diagram {
...
     foreach my $tweak (@$tweaks) {
-        print "\nTweak.Request\t$tweak\n";
+        push @response, "\nTweak.Request\t$tweak\n";
 
         if($tweak=~/^pipeline\.param\[(\w+)\](\?|#|=(.+))$/) {
             my ($param_name, $operator, $new_value_str) = ($1, $2, $3);
-
+            my $tweakStructure;
+            $tweakStructure->{Action} = $self->{ACTION}->{substr($operator, 0, 1)};
+            $tweakStructure->{Object}->{Type} = 'Pipeline';
```
muffato: What about defining all the object types in a separate hash like the error messages and the actions ? This would guarantee consistency across the various regexp each type accepts

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -715,43 +768,53 @@ sub apply_tweaks {
...
                 my $analysis_name = $analysis->logic_name;
-
+                my $tweakStructure;
+                $tweakStructure->{Object}->{Type} = 'Analysis';
+                $tweakStructure->{Object}->{Id} = $analysis->dbID + 0;
+                $tweakStructure->{Object}->{Name} = $analysis_name;
+                $tweakStructure->{Action} = $self->{ACTION}->{substr($operator, 0, 1)};
                 if( $attrib_name eq 'resource_class' ) {
+                    $tweakStructure->{Return}->{OldValue} = $analysis->resource_class ? $analysis->resource_class->name : undef;
+                    $tweakStructure->{Return}->{Field} = 'resource_class';
```
muffato: Set `$tweakStructure->{Return}->{Field}` to `$attrib_name` before the `if`, which would cover all cases

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +825,81 @@ sub apply_tweaks {
                     }
 
                 } elsif( $attrib_name eq 'is_excluded' ) {
+                    $tweakStructure->{Return}->{Field} = 'is_excluded';
```
muffato: Don't need this if `$tweakStructure->{Return}->{Field}` is set above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +825,81 @@ sub apply_tweaks {
...
                         } else {
-                           print "Tweak.Changing\tanalysis[$analysis_name].is_excluded ::\t" .
+                           push @response, "Tweak.Changing\tanalysis[$analysis_name].is_excluded ::\t" .
                                $analysis_stats->is_excluded() . " --> $new_value_str\n";
                            $analysis_stats->is_excluded($new_value);
                            $need_write = 1;
                         }
                     }
                 } elsif($analysis->can($attrib_name)) {
+                    $tweakStructure->{Return}->{Field} = $attrib_name;
```
muffato: Don't need this if `$tweakStructure->{Return}->{Field}` is set above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -762,57 +825,81 @@ sub apply_tweaks {
...
-                        print "Tweak.Changing\tanalysis[$analysis_name].$attrib_name ::\t$old_value --> ".stringify($new_value)."\n";
-
+                        $tweakStructure->{Return}->{NewValue} = stringify($new_value);
+                        push @response, "Tweak.Changing\tanalysis[$analysis_name].$attrib_name ::\t$old_value --> ".stringify($new_value)."\n";
                         $analysis->$attrib_name( $new_value );
                         $need_write = 1;
                     }
                 } else {
-                    print "Tweak.Error   \tAnalysis does not support '$attrib_name' attribute\n";
+                    $tweakStructure->{Return}->{Field} = $attrib_name;
```
muffato: Don't need this if `$tweakStructure->{Return}->{Field}` is set above

t/03.scripts/tweak_pipeline_output.t
```diff
@@ -0,0 +1,113 @@
...
+        ok(exists($tweakJson->{Return}->{NewValue}) || $tweakJson->{Warning} || $tweakJson->{Error}, 'New value exists for ' . join (' ', @{$request}));
```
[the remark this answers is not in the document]
mira13: exclusive ORs are not applicable, 
Here examples where we have both error and value (Thera are more cases at all):

 -tweak analysis[perform_cmd].is_excluded=20 -json

 $VAR1 = [
          {
            'Error' => 'Invalid value',
            'Action' => 'SET',
            'Object' => {
                           'Type' => 'Analysis',
                           'Id' => 1,
                           'Name' => 'perform_cmd'
                         },
             'Return' => {
                           'Field' => 'is_excluded',
                           'OldValue' => 0,
                           'NewValue' => '20'
                         }
           }
         ];



 - Tweaks response received for -tweak analysis[perform_cmd].is_excluded\# -json

 $VAR1 = [
           {
             'Error' => 'Action is not supported',
             'Return' => {
                           'OldValue' => 0,
                           'Field' => 'is_excluded'
                         },
             'Action' => 'DELETE',
             'Object' => {
                           'Id' => 1,
                           'Name' => 'perform_cmd',
                           'Type' => 'Analysis'
                         }
           }
         ];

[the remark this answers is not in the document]
muffato: OK, no _xor_ then ...

mira13:
@muffato Review comments resolved.

ens-bwalts:
@muffato 's fix has been merged up to master. Can you rebase these onto master to make sure that there aren't any conflicts and that everything still works.

Also, could you squash some of the commits to give a cleaner commit history (e.g. some of the syntax fixes could be squashed into a single commit)

muffato:
Yes, cleaning up the commit history is the next important bit to address. Each commit should do 1 thing and have an informative message. I've asked to restore the indentation to make it clear when we diff the files what has really changed (BTW there are still a few lines in HivePipeline.pm that need to be fixed). But I also want to avoid commits undoing a previous commit to restore the original code, as it fools `git annotate` and such. 
Of course we could squash all your commits into one when accepting your pull-request, and the history would still be available here but perhaps you can do the clean up on your side ?

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -841,16 +937,25 @@ sub apply_tweaks {
                             'submission_cmd_args'   => $new_submission_cmd_args,
                             'worker_cmd_args'       => $new_worker_cmd_args,
                         );
+                        $tweakStructure->{Return}->{OldValue} = undef;
                     }
+                    $tweakStructure->{Return}->{Field} = $meadow_type;
+                    $tweakStructure->{Return}->{NewValue} = stringify([$new_submission_cmd_args, $new_worker_cmd_args]);
                     $need_write = 1;
+                    push @{$responseStructure->{Tweaks}}, $tweakStructure;
```
muffato: Move this line before `$need_write`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +564,187 @@ sub print_diagram {
...
                             $cr_collection->forget_and_mark_for_deletion( $c_rule );
                             $need_write = 1;
 
-                            print "Tweak.Deleting\t".$c_rule->toString." --> (missing value)\n";
+                            push @response, "Tweak.Deleting\t".$c_rule->toString." --> (missing value)\n";
                         }
                     }
 
                     if($operator eq '=' or $operator eq '+=') {     # create new rules
+                        $tweakStructure->{Return}->{NewValue} = $tweakStructure->{Return}->{OldValue} . $new_value;
```
muffato: _OldValue_ is an array-ref, not a string. `$new_value` can be a string or an array-ref

mira13:
@muffato All fixed

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +564,183 @@ sub print_diagram {
...
+                $tweakStructure->{Return}->{Field} = $param_name;
+                $tweakStructure->{Return}->{OldValue} =  exists($param_hash->{ $param_name }) ? stringify($param_hash->{ $param_name }) : undef;
 
                 if($operator eq '?') {
-                    print "Tweak.Show    \tanalysis[$analysis_name].param[$param_name] ::\t"
-                        . (exists($param_hash->{ $param_name }) ? stringify($param_hash->{ $param_name }) : '(missing value)')
-                        ."\n";
+                    $tweakStructure->{Return}->{NewValue} = $tweakStructure->{Return}->{OldValue};
+                    push @response, "Tweak.Show    \tanalysis[$analysis_name].param[$param_name] ::\t"
+    	                . (exists($param_hash->{ $param_name }) ? stringify($param_hash->{ $param_name }) : '(missing value)')
```
muffato: Still not quite right :(

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +564,183 @@ sub print_diagram {
...
+	                    ."\n";
```
muffato: Still not quite right :(

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +564,183 @@ sub print_diagram {
...
+            $tweakStructure->{Object}->{Name} = undef;
+            $tweakStructure->{Return}->{Field} = $param_name;
+            $tweakStructure->{Return}->{OldValue} = $value;
 
             if($operator eq '?') {
-                print "Tweak.Show    \tpipeline.param[$param_name] ::\t"
-                     . ($hash_pair ? $hash_pair->{'param_value'} : '(missing_value)') . "\n";
+                $tweakStructure->{Return}->{NewValue} = $value;
+                push @response, "Tweak.Show    \tpipeline.param[$param_name] ::\t"
+	                . ($hash_pair ? $hash_pair->{'param_value'} : '(missing_value)') . "\n";
```
muffato: Still not quite right :(

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +564,183 @@ sub print_diagram {
...
     foreach my $tweak (@$tweaks) {
-        print "\nTweak.Request\t$tweak\n";
+        push @response, "\nTweak.Request\t$tweak\n";
 
         if($tweak=~/^pipeline\.param\[(\w+)\](\?|#|=(.+))$/) {
             my ($param_name, $operator, $new_value_str) = ($1, $2, $3);
-
             my $pwp_collection  = $self->collection_of( 'PipelineWideParameters' );
             my $hash_pair       = $pwp_collection->find_one_by('param_name', $param_name);
+            my $value = $hash_pair ? $hash_pair->{'param_value'} : undef;
```
muffato: Align the equal sign with the lines above

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +564,183 @@ sub print_diagram {
...
                             $need_write = 1;
 
-                            print "Tweak.Deleting\t".$c_rule->toString." --> (missing value)\n";
+                            push @response, "Tweak.Deleting\t".$c_rule->toString." --> (missing value)\n";
                         }
                     }
 
                     if($operator eq '=' or $operator eq '+=') {     # create new rules
+                        my $new_value_str = ref($new_value) eq 'ARRAY' ? join(' ', $new_value) : $new_value;
+                        $tweakStructure->{Return}->{NewValue} = join(' ', $tweakStructure->{Return}->{OldValue}) . $new_value_str;
```
muffato: _OldValue_ is a list. _NewValue_ should be a list too

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: Sorry, something was wrong with my editor, it shows in another way tan it is in reality. Now checked in different editor and can see the difference.

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +564,183 @@ sub print_diagram {
...
+            $tweakStructure->{Object}->{Name} = undef;
+            $tweakStructure->{Return}->{Field} = $param_name;
+            $tweakStructure->{Return}->{OldValue} = $value;
 
             if($operator eq '?') {
-                print "Tweak.Show    \tpipeline.param[$param_name] ::\t"
-                     . ($hash_pair ? $hash_pair->{'param_value'} : '(missing_value)') . "\n";
+                $tweakStructure->{Return}->{NewValue} = $value;
+                push @response, "Tweak.Show    \tpipeline.param[$param_name] ::\t"
+                    . ($hash_pair ? $hash_pair->{'param_value'} : '(missing_value)') . "\n";
```
muffato: Don't think I'm enjoying it, but this is still not right

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,148 +564,183 @@ sub print_diagram {
...
+                $tweakStructure->{Return}->{OldValue} =  exists($param_hash->{ $param_name }) ? stringify($param_hash->{ $param_name }) : undef;
 
                 if($operator eq '?') {
-                    print "Tweak.Show    \tanalysis[$analysis_name].param[$param_name] ::\t"
-                        . (exists($param_hash->{ $param_name }) ? stringify($param_hash->{ $param_name }) : '(missing value)')
-                        ."\n";
+                    $tweakStructure->{Return}->{NewValue} = $tweakStructure->{Return}->{OldValue};
+                    push @response, "Tweak.Show    \tanalysis[$analysis_name].param[$param_name] ::\t"
+                    . (exists($param_hash->{ $param_name }) ? stringify($param_hash->{ $param_name }) : '(missing value)')
+                    ."\n";
```
muffato: Still not... Have a look at `version/2.5`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
mira13: Before my changes intentions were different here and on line 669 (5 here and 4 there), rather confusing. On line 669 you commented to do is it was before. Should I do 4 or 5 spaces???

mira13:
@muffato 
1. Intention changed for 4 spaces, its is not like was before, but I assume consistent
2. Old value print fixed

muffato:
Of the three lines that didn't have the right indentation, the first one is the only one left to fix. It is missing 1 space

My point was not to change _OldValue_ (we said some time ago that it should be a list) but to make _NewValue_ a list too rather than a string. Also, note that some of the new values could be redundant with the old ones. In `+=` mode, eHive will detect those and make a unique list of analyses. The accurate way of getting the new values is not to append the user input `$new_value`, but to look in the collection again.

muffato:
Can you please rebase this branch on top of master too ?

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +564,184 @@ sub print_diagram {
...
                         }
                     }
 
                     if($operator eq '=' or $operator eq '+=') {     # create new rules
                         Bio::EnsEMBL::Hive::Utils::PCL::parse_wait_for($self, $analysis, $new_value);
-                        foreach my $c_rule ( @{$analysis->control_rules_collection} ) {
-                            print "Tweak.Adding\t".$c_rule->toString."\n";
-                        }
+                        my $new_analyses = $self->collection_of( 'Analysis' )->find_one_by( 'dbID' => $analysis->dbID) ;
+                        my $acr_collection  = $new_analyses->control_rules_collection;
```
muffato: There is already `$analysis`

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +564,184 @@ sub print_diagram {
...
+                $tweakStructure->{Action} = $self->{TWEAK_ACTION}->{substr($operator, 0, 1)};
+                $tweakStructure->{Object}->{Id} = $analysis->dbID + 0;
+                $tweakStructure->{Object}->{Name} = $analysis_name;
+                $tweakStructure->{Return}->{Field} = $attrib_name;
                 if( $attrib_name eq 'wait_for' ) {
-
                     my $cr_collection   = $self->collection_of( 'AnalysisCtrlRule' );
                     my $acr_collection  = $analysis->control_rules_collection;
-
+                    $tweakStructure->{Return}->{OldValue} = join (' ', map { $_->condition_analysis_url } @$acr_collection);
```
muffato: Change this back to being a list, not a string

modules/Bio/EnsEMBL/Hive/HivePipeline.pm
```diff
@@ -544,151 +564,184 @@ sub print_diagram {
...
                     }
 
                     if($operator eq '=' or $operator eq '+=') {     # create new rules
                         Bio::EnsEMBL::Hive::Utils::PCL::parse_wait_for($self, $analysis, $new_value);
-                        foreach my $c_rule ( @{$analysis->control_rules_collection} ) {
-                            print "Tweak.Adding\t".$c_rule->toString."\n";
-                        }
+                        my $new_analyses = $self->collection_of( 'Analysis' )->find_one_by( 'dbID' => $analysis->dbID) ;
+                        my $acr_collection  = $new_analyses->control_rules_collection;
+                        $tweakStructure->{Return}->{NewValue} = join (' ', map { $_->condition_analysis_url } @$acr_collection);
```
muffato: We said this needs to be a list, not a string

mira13:
Fixed, waiting for review

muffato:
I've added a few commits to correct the indentation and introduce extra messages that tweak_pipeline should print an have only recently been added to master

muffato:
:confetti_ball:

PR #63, opened 2019-01-24
Collection field regex search

tgrego:
This is a follow up to https://github.com/Ensembl/ensembl-hive/pull/61.
It relates to https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-2558 but the scope has increased to any field and not just Analysis tags.

A new `~` search operator is introduced, that allow regex searches on named fields.
It extends the Analyses pattern syntax with `expression = <word> '~' <any>` where any can be a regex (contain `%`).

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Utils/Collection.pm
+++ b/modules/Bio/EnsEMBL/Hive/Utils/Collection.pm
@@ -206,7 +206,7 @@ sub _find_all_by_subpattern {    # subpatterns can be combined into full pattern
         my $field = $1;
         $pattern = $2;
         $pattern =~ s/\%/.*/g;
-        $filtered_elements = $self->find_all_by( $field, sub { return $_[0 ]=~ /^.*,?${pattern},?.*$/; } );
+        $filtered_elements = $self->find_all_by( $field, sub { return $_[0 ]=~ /^(.*,)?${pattern}(,.*)?$/; } );
 
     } else {
         die "The pattern '$pattern' is not recognized\n";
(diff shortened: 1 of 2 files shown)


muffato:
Hi Tiago,
Yes, that looks good, and the test is clear too. However you've done more than just adding the `~` operator: you made it support array-ref values too. Is that needed ? Do we have such fields? If we do, I would suggest to split your commit into two: one that implements `~` for scalars, and one to support array-refs in _all_ operators for consistency.
Then, specifically about your implementation, you've chosen to return true when _at least one_ of the values matches the search term. You could have made it return true when _all_ the values would have to match. I think you've chosen the most sensible implementation but for other operators, does it make sense to do the same ? Or what about offering both options, e.g. `any(colour==red)`, `all(marks>4)` ? Ideally, this would be driven by a need to filter existing fields.
Otherwise, if we don't need the functionality, we might not add it at all to keep the code lean

tgrego:
I did not technically test the code with collections of Analysis (only the array collection in the tests) but from what I see in the Analysis code, the return value of `$analysis->tags` is an array, thus array support is required if we want this to work for the analysis tags.
My ehive experience is however limited so I might be wrong!

tgrego:
Took a second look at the code and documentation, and now it looks like Analysis tags can take either an array or a scalar (string of comma separated tags). The actual code should be happy with whatever is passed on at `$analysis->new` time, there does not seem to be an enforcement of any kind...

muffato:
(assuming Travis is happy)

PR #64, opened 2019-01-24
Experimental/add job semaphore view

mira13:
New view for job-semaphore


muffato:
FYI, #44 already has two changes of mine related to views. The idea was to bundle them with @ens-bwalts' new semaphore views in order to have just 1 schema version bump and 1 patch. Once we're happy with the semaphore views (from this PR or #44), can we still do that ?

Re. the change itself, yes this view mimics the previous `job` table. I think that's exactly what people used to the version 2.4 would like. Note that you'll have to remove the references to `my_pipeline` and extend `sql/procedures.*`. Also, are the `ALGORITHM`, `root@localhost` etc lines needed ? All our current views start with `CREATE OR REPLACE VIEW` only.

ens-bwalts:
This view is simpler than what I was trying to do in #44 , and is probably all that we need to combine semaphore and job information into one table. I'm happy to use this in place of my attempt in that PR.

muffato:
Note that all these commits will have to be squashed into 1 and a schema version tag created

tgrego:
offtopic @mira13: Why do you commit with user 'mira' but have 'mira13' username in github?
it's usually nicer to use just a username (plus this is the real @mira in github).

muffato:
Sorry but the patches then need to patch the `drop_hive_tables` procedure too

mira13:
@muffato 
> Need to add semaphore_job to drop_hive_tables

resolved

muffato:
I've found some other issues:
 - `semaphore_job` is implemented with a LEFT JOIN in PostreSQL, but a regular JOIN in the other drivers. Can you make them the same ?
- In `procedures.mysql`, `semaphore_job` is placed after `live_roles`, whereas it is before `resource_usage_stats` in the offer ones
In eHive, I find it best practice to make the per-driver files as consistent as possible, so that you can _diff_ them and quickly spot the differences (which should only be due to having different SQL flavours)

mira13:
>I've found some other issues:

@muffato 
Thanks, it's resolved now, all views are now before _resource_usage_stats_ , Left join replaced with join.

PR #66, opened 2019-02-01
Feature/reset job by input id

s-mm:
Reset a job with given input_id and analyses pattern. Both options can be wildcard arguments.


modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,61 @@ sub reset_or_grab_job_by_dbID {
...
+    my $sth = $self->prepare( $sql );
+    my $return_code = $sth->execute( @values )
+        or die "Could not run\n\t$sql\nwith data:\n\t(".join(',', @values).')';
+    $sth->finish;
+
+
+    # Get list of input_id for given wildcard argument
+    my $sql_get_input_id = qq{
+        SELECT input_id FROM job
+         WHERE input_id LIKE ? AND analysis_id=?
```
ens-bwalts: This won't work because input_id is subject to overflow. That is to say that the input_id value might be stored in the input_id column, or it might be stored in the analysis_data table. If the input_id is stored in analysis_data, then input_id will have a value like "_extended_data_id N" (where N is an arbitrary integer). See fetch_by_analysis_id_and_input_id for an example.

One approach here would be to fetch_all_by_analysis_id, which will get a list of relevant jobs. These jobs can then be filtered by input_id. If you call input_id on an AnalysisJob object, the value will be correct. BaseAdaptor automatically checks to see if the input_id is stored in the input_id column or has been overflow-ed to analysis_data and fetches it appropriately. Eventually, you will get a set of jobs, and you can reset them by job_id.

muffato: Fetching all jobs is going to be a killer: there can be millions of jobs. If you want to go this way, it might be worth it to add a new `fetch_iterator` in DBSQL::BaseAdaptor and process them on the fly.
Another thing: job parameters can be available through their "stack" too (`accu_id_stack`, `param_id_stack`)

ens-bwalts: The issue of job count is difficult to get around due to overflow. I'm not coming up with a better idea than an iterator, which at least solves the memory overhead of storing potentially millions of jobs. A view joining analysis_data will put even more load on the DB.

As for the second part, I'm inclined to have this feature just scan input_id in the first instance. It would be more useful to be able to select jobs based on any visible parameter. Adding that feature would pair nicely with adding a parameter inspector for jobs, so I'm inclined to add that in a second iteration.

muffato: Re. the stack-inherited parameters: you'll have to be clear in the documentation that the feature doesn't work on pipelines that have the stack enabled at the pipeline level, or on analyses that follow `INPUT_PLUS`.

ens-bwalts:
After discussion, we will leave the first iteration of this simple - find by input_id, and go ahead and accept the performance penalty of a fetch_all_... rather than implementing an iterator. If and when usage justifies further work we will improve the solution

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,61 @@ sub reset_or_grab_job_by_dbID {
...
+               The retry_count will be set to 1 for previously run jobs (partially or wholly) to trigger PRE_CLEANUP for them,
+               but will not change retry_count if a job has never *really* started. $input_id can be a wildcard entry.
+  Returntype : Bio::EnsEMBL::Hive::AnalysisJob or undef
+
+=cut
+
+sub reset_or_grab_job_by_inputID {
+    my ($self, $analysis_id, $input_id, $role_id) = @_;
+
+    my $new_status  = $role_id ? 'CLAIMED' : 'READY';
```
muffato: `role_id` seems to be never given. Remove the option ?

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,61 @@ sub reset_or_grab_job_by_dbID {
...
+        or die "Could not run\n\t$sql\nwith data:\n\t(".join(',', @values_get_input_id).')';
+    my $data_get_input_id = $sth_get_input_id->fetchrow_array();
+    if (! $data_get_input_id) {
+        die "Could not find input_id for given input_id and analysis_pattern";
+    }
+    my $job = $self->fetch_by_analysis_id_and_input_id($analysis_id, $data_get_input_id);
+    $sth_get_input_id->finish;
+
+    return $job;
+}
```
muffato: Semaphores need to be updated as well. Have a look at `reset_jobs_for_analysis_id` to see how it's done (it's tricky)

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,61 @@ sub reset_or_grab_job_by_dbID {
...
+        SELECT input_id FROM job
+         WHERE input_id LIKE ? AND analysis_id=?
+    };
+
+    my @values_get_input_id = ($input_id, $analysis_id);
+
+    my $sth_get_input_id = $self->prepare( $sql_get_input_id ) or die "Unable to prepare" . $self->errstr;
+    my $return_code_input_id = $sth_get_input_id->execute(@values_get_input_id)
+        or die "Could not run\n\t$sql\nwith data:\n\t(".join(',', @values_get_input_id).')';
+    my $data_get_input_id = $sth_get_input_id->fetchrow_array();
```
muffato: Would be more efficient if you get the `job_id` in the query

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,61 @@ sub reset_or_grab_job_by_dbID {
...
+    if (! $data_get_input_id) {
+        die "Could not find input_id for given input_id and analysis_pattern";
+    }
+    my $job = $self->fetch_by_analysis_id_and_input_id($analysis_id, $data_get_input_id);
```
muffato: There can be more than 1 job that match the input_id pattern

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
muffato: `reset_jobs_for_analysis_id` does a bulk update of many jobs at once. That's why it's got the various big SQL to select and update the rows and the transaction to ensure consistency. This is not very fitting here because you call this method for each job independently.

You could maybe in a first implementation just copy `specialize_worker` from Queen (the part where `$job_id` is defined). The code there is clearer and only uses the API to fetch the semaphore and update it.

Because that could potentially update the same semaphore many times, it'd then be nice (in a second implementation) to buffer and aggregate the semaphore updates to apply them at the end. But then you may still need some lock / transaction mechanisms to ensure consistency

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
...
+=cut
+
+sub reset_job_by_input_id_and_sync {
+    my ($self, $input_id_pattern, @analyses_list) = @_;
+    $input_id_pattern =~ s/\%/\.*/g;
+    my $analysis;
+    my $jobs = $self->fetch_all_by_analysis_id_status(@analyses_list);
+    foreach my $job (@$jobs){
+         my $input_id = $job->input_id;
+         if ($input_id =~ /$input_id_pattern/) {
```
muffato: What about pre-compiling the regexp with `qr` ? According to the doc it would be faster.

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
...
+    Reset jobs for the specified $input_id pattern and analyses_id pattern. $input_id can be a wildcard argument.
+  Returntype : none
+  Exceptions :
+  Caller     : beekeeper.pl
+
+=cut
+
+sub reset_job_by_input_id_and_sync {
+    my ($self, $input_id_pattern, @analyses_list) = @_;
+    $input_id_pattern =~ s/\%/\.*/g;
```
muffato: The pattern you currently accept is in fact any Perl regexp, with the addition of substituting `%` with `.*` to give it a MySQL taste. However, `a%b` means "starts with `a` and ends with `b`" in MySQL, but "contains an `a` and then a `b` in Perl. Is that fine ? @ens-bwalts ?

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
...
+    my $analysis;
+    my $jobs = $self->fetch_all_by_analysis_id_status(@analyses_list);
+    foreach my $job (@$jobs){
+         my $input_id = $job->input_id;
+         if ($input_id =~ /$input_id_pattern/) {
+                my $job_status = $job->status();
+
+                if($job_status =~ $ALL_STATUSES_OF_RUNNING_JOBS) {
```
muffato: Does this work ?? `$ALL_STATUSES_OF_RUNNING_JOBS` is a string
In Queen I've done
```if ($Bio::EnsEMBL::Hive::DBSQL::AnalysisJobAdaptor::ALL_STATUSES_OF_TAKEN_JOBS =~ /'$job_status'/) {```

File: modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm [hunk 9]
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
...
+                       die "Job is already in progress, cannot reset";
```
muffato: Add information about the job. _AnalysisJob_ may have a `toString` method

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
...
+                }
+
```
muffato: At this stage, jobs are either `CLAIMED`, `DONE`, `PASSED_ON`, `FAILED`, `SEMAPHORED` or `READY`

`CLAIMED` and `SEMAPHORED` can't be reset. `READY` doesn't need to be reset

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
...
+                if(($job_status eq 'DONE') and my $controlled_semaphore = $job->controlled_semaphore) {
+                       $controlled_semaphore->increase_by( [ $job ] );
+                }
+                $analysis = $job->analysis;
+                $job->set_and_update_status('READY');
+                $analysis->stats->adaptor->increment_a_counter( $Bio::EnsEMBL::Hive::AnalysisStats::status2counter{$job->status}, 1, $job->analysis_id );
```
muffato: Should decrement the counter of the previous job status too

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
...
+
+                # Syncing stats. Copied from synchronize_AnalysisStats()
+                my $stats = $job->analysis->stats;
+                $stats->refresh();
+
+                my $job_counts = $stats->hive_pipeline->hive_use_triggers() ? undef : $self->fetch_job_counts_hashed_by_status( $stats->analysis_id );
+
+                $stats->recalculate_from_job_counts( $job_counts );
+
+                $stats->update;
```
muffato: I think this section (since setting `$stats`) is not needed since you already take care of the counters above (once you decrement the other one)

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
     return $job;
 }
 
+=head2 reset_job_by_input_id_and_sync
+
+  Arg [1]: string $input_id pattern
+  Arg [2]: array $analysis_id
```
muffato: The array must contain _Analysis_ objects

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
...
+  Example:
+    my $job = $self->{'dba'}->get_AnalysisJobAdaptor->reset_job_by_input_id_and_sync($input_id_pattern, $analyses_pattern);
```
muffato: This method gets the _Analysis_ objects themselves, not the pattern

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,54 @@ sub reset_or_grab_job_by_dbID {
...
+  Description:
+    Reset jobs for the specified $input_id pattern and analyses_id pattern. $input_id can be a wildcard argument.
```
muffato: Same comment about the analysis pattern

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
ens-bwalts: To be consistent with other pattern matching (such as -analyses_pattern in runWorker.pl), this should be MySql style - so a%b should be starts with a and ends with b, and if the user wants to match any leading string, it would need to be %a%b

scripts/beekeeper.pl
```diff
@@ -71,6 +71,7 @@ sub main {
     my $unblock_semaphored_jobs     = 0;    # Mark SEMAPHORED jobs to READY
     my $forgive_failed_jobs         = 0;    # Mark FAILED jobs to DONE
     my $discard_ready_jobs          = 0;    # Mark READY jobs to DONE
+    my $reset_job_for_input_id      = 0;
```
muffato: It should probably be `jobs`, not `job`. Would be good to clarify which statuses are changed and to what.

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,59 @@ sub reset_or_grab_job_by_dbID {
...
+         $input_id_pattern = qr/$input_id_pattern/;
+         if ($input_id =~ /$input_id_pattern/) {
+                my $job_status = $job->status();
+                my $job_info = $job->toString;
+                if($job_status =~ /CLAIMED|SEMAPHORED/) {
+                       die "$job_info is $job_status, cannot reset";
+                }
+
+                if ($job_status ne 'READY') {
+                   if(($job_status eq 'DONE') and my $controlled_semaphore = $job->controlled_semaphore) {
```
muffato: `PASSED_ON` is like `DONE`: the job is considered "finished" and it has decremented the semaphore counter

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,59 @@ sub reset_or_grab_job_by_dbID {
...
+    my $analysis;
+    my $jobs = $self->fetch_all_by_analysis_id_status(@analyses_list);
+    foreach my $job (@$jobs) {
+         my $input_id = $job->input_id;
+         $input_id_pattern = qr/$input_id_pattern/;
+         if ($input_id =~ /$input_id_pattern/) {
+                my $job_status = $job->status();
+                my $job_info = $job->toString;
+                if($job_status =~ /CLAIMED|SEMAPHORED/) {
+                       die "$job_info is $job_status, cannot reset";
```
muffato: I'm not sure the code should _die_ here, maybe simply skip the job ?
This would definitely be the correct behaviour for "reset by job_id", i.e. "Can you reset job_id 1234 for me ? No, I can't, it's CLAIMED".
But what do you think the question is here ?
1. "Can you reset _all_ the jobs that have `species => human` for me ?" -> Should die if at least 1 job can't be reset, probably including jobs that are running
1. "Can you reset all the _resettable_ jobs that have `species => human` for me ?" Should silently skip the jobs that are already running / claimed or can't be reset.

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,59 @@ sub reset_or_grab_job_by_dbID {
...
+                }
+
+                if ($job_status ne 'READY') {
```
muffato: Since you have removed the line with `ALL_STATUSES_OF_RUNNING_JOBS`, this will now catch jobs that are running. These should _not_ be reset.
The test you had in  previous commit: https://github.com/Ensembl/ensembl-hive/blob/6f07a94bfe737c5911bf24dab06c13588af8e153/modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm#L614-L616 was fine and should be reinstated, but now taking into account my new comment: should it _die_ or _skip_ ?

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
muffato: :+1:

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,59 @@ sub reset_or_grab_job_by_dbID {
...
+sub reset_job_by_input_id_and_sync {
+    my ($self, $input_id_pattern, @analyses_list) = @_;
+
+    #Convert MySQL wildcard pattern to Perl wildcard pattern
+    if ($input_id_pattern =~ /\A[^%]/) {
+        $input_id_pattern = '^'.$input_id_pattern;
+    }
+    if ($input_id_pattern =~ /[^%]\z/) {
+        $input_id_pattern = $input_id_pattern.'$';
+    }
```
muffato: See https://github.com/Ensembl/ensembl-hive/blob/version/2.5/modules/Bio/EnsEMBL/Hive/Utils/Collection.pm#L176-L179 which implements "analyses pattern" with `%` too.
 `^` and `$` are added regardless of whether the requested pattern starts or ends with `%`. I don't know if it makes a difference in terms of performance: your solution is certainly more accurate but we could just not bother and _always_ add `^` and `$` in the `qr` expression, and save 6 lines of code.
Or should we change Collection.pm ?
Or factor this out in Utils.pm ? Add there a simple function that takes a user pattern and returns the perl `qr` object, which could be used in both this file and Collection.pm . Actually, I like this idea ! It would make it easier to extend the language if we ever want to !

scripts/beekeeper.pl
```diff
@@ -284,6 +286,15 @@ sub main {
 
     if($reset_job_id) { $queen->reset_job_by_dbID_and_sync($reset_job_id); }
 
+    if($reset_job_for_input_id and $self->{'analyses_pattern'}) {
+        my @analyses_list = $self->{'pipeline'}->collection_of('Analysis')->find_all_by_pattern( $self->{'analyses_pattern'});
```
muffato: `find_all_by_pattern` returns an array-_ref_. Actually, I'd also prefer passing a reference to `reset_job_by_input_id_and_sync` rather than the expanded list. It makes it easier to add extra arguments if we need to

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,57 @@ sub reset_or_grab_job_by_dbID {
...
+         if ($input_id =~ /$input_id_pattern/) {
+                my $job_status = $job->status();
+                my $job_info = $job->toString;
+                if($job_status =~ /$ALL_STATUSES_OF_RUNNING_JOBS|CLAIMED|SEMAPHORED/) {
+                   warn "$job_info is $job_status, cannot reset - skipping to next job";
+                   next;
+                }
+
+                #The job at this stage will be DONE, PASSED_ON, FAILED or READY
+                if ($job_status ne 'READY') {
```
muffato: The negative test here is a bit strange. The previous `if` follows the paradigm "detect and process the special cases early". READY jobs are the same: print a warning and move on. I would like the same approach: if the job is READY, `warn` and `next`.

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,57 @@ sub reset_or_grab_job_by_dbID {
...
+    #Convert MySQL wildcard pattern to Perl wildcard pattern
+    $input_id_pattern = return_qr_pattern($input_id_pattern);
+
+    my $analysis;
+    my $jobs = $self->fetch_all_by_analysis_id_status($analyses_list);
+    foreach my $job (@$jobs) {
+         my $input_id = $job->input_id;
+         if ($input_id =~ /$input_id_pattern/) {
+                my $job_status = $job->status();
+                my $job_info = $job->toString;
```
muffato: Align the `=` sign

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,60 @@ sub reset_or_grab_job_by_dbID {
...
+    foreach my $job (@$jobs) {
+        my $input_id = $job->input_id;
+        if ($input_id =~ /$input_id_pattern/) {
+            my $job_status = $job->status();
+            my $job_info   = $job->toString;
+            if($ALL_STATUSES_OF_RUNNING_JOBS =~ /'$job_status'/ | $job_status =~ /CLAIMED|SEMAPHORED/) {
+                warn "$job_info is $job_status, cannot reset - skipping to next job";
+                next;
+            }
+
```
muffato: Can you remove the trailing whitespace ?

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,60 @@ sub reset_or_grab_job_by_dbID {
...
+                }
+                if(($job_status eq 'PASSED_ON') and my $controlled_semaphore = $job->controlled_semaphore) {
+                    $controlled_semaphore->decrease_by( [ $job ] );
+                }
+                $analysis = $job->analysis;
+                $analysis->stats->adaptor->increment_a_counter( $Bio::EnsEMBL::Hive::AnalysisStats::status2counter{$job->status}, -1, $job->analysis_id );
+                $job->set_and_update_status('READY');
+                $analysis->stats->adaptor->increment_a_counter( $Bio::EnsEMBL::Hive::AnalysisStats::status2counter{$job->status}, 1, $job->analysis_id );
+            }
+
```
muffato: Can you remove the trailing whitespace ?

modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
```diff
@@ -587,6 +587,60 @@ sub reset_or_grab_job_by_dbID {
...
+                next;
+            }
+
+            # The job at this stage will be DONE, PASSED_ON, FAILED
+            else {
+                if(($job_status eq 'DONE') and my $controlled_semaphore = $job->controlled_semaphore) {
+                    $controlled_semaphore->increase_by( [ $job ] );
+                }
+                if(($job_status eq 'PASSED_ON') and my $controlled_semaphore = $job->controlled_semaphore) {
+                    $controlled_semaphore->decrease_by( [ $job ] );
```
muffato: Should be `increase_by` here too. Like `DONE` the job was considered finished

modules/Bio/EnsEMBL/Hive/Utils.pm
```diff
@@ -505,5 +505,25 @@ sub print_aligned_fields {
 }
 
 
+=head2 return_qr_pattern
+
+    Argument[0]: Wildcard pattern
+    Description: For the given pattern, the method will return the qr object pattern.
```
muffato: Can you remove the trailing whitespace on these three lines ?

muffato:
Thank you !

scripts/beekeeper.pl
```diff
@@ -286,6 +288,15 @@ sub main {
 
     if($reset_job_id) { $queen->reset_job_by_dbID_and_sync($reset_job_id); }
 
+    if($reset_jobs_for_input_id and $self->{'analyses_pattern'}) {
```
ens-bwalts: This should be moved down to be part of the "has_task" group at line 382. This is because that's after `@$list_of_analyses `has been built and checked to be sure that it is not empty. Otherwise, if the user provides an analyses_pattern that doesn't match anything, an empty list of analyses eventually makes its way to `AnalysisJobAdaptor::fetch_all_by_analysis_id_status()`, which causes a crash.

t/03.scripts/beekeeper_reset.t
```diff
@@ -134,6 +134,10 @@ foreach my $pipeline_url (@$ehive_test_pipeline_urls) {
     beekeeper($hive_url, ['-discard_ready_jobs'], 'beekeeper.pl -discard_ready_jobs');
     assert_jobs($job_adaptor, [['DONE',1,0],['READY',0,0],['DONE',1,0],['DONE',1,0],['DONE',1,0],['DONE',1,0],['DONE',1,0]] );
 
+    # Reset a job with specific input_id and analysis_pattern
+    beekeeper($hive_url, ['-reset_jobs_for_input_id', '%640%', '-analyses_pattern', 2], 'beekeeper.pl -reset_jobs_for_input_id -analyses_pattern');
```
ens-bwalts: This tests that it doesn't reset a job if the input_id does not match any job. There should be another test that checks that a job gets reset if the input_id does match. For example

`beekeeper($hive_url, ['-reset_jobs_for_input_id', '%4%', '-analyses_pattern', 2], 'beekeeper.pl -reset_jobs_for_input_id -analyses_pattern');`

Which should give expected jobs of
`[['DONE',1,0],['READY',0,0],['DONE',1,0],['DONE',1,0],['DONE',1,0],['DONE',1,0],['READY',1,0]]`

sgiorgetti:
Decided not to go on with this PR.
Closing

PR #67, opened 2019-02-04
Add a beekeeper option for attempting to stop the whole pipeline

mkszuba:
Block all active Beekeepers, attempt to kill all active workers, report on both. This can fail to shut everything down because the current beekeeper might not be able to reach certain meadows (_e.g._ LOCAL on a different host) or have insufficient permissions to kill certain workers (_e.g._ processes or batch jobs belonging to a different user) but it seems to be all we could do without actively communicating with other Beekeepers and it's better than nothing.

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Beekeeper.pm
+++ b/modules/Bio/EnsEMBL/Hive/Beekeeper.pm
@@ -187,5 +187,27 @@ sub check_if_blocked {
     return $self->is_blocked;
 }
 
+
+=head2 toString
+
+  Example     : print $beekeeper->toString();
+  Description : Produces a string summary of properties of this beekeeper.
+  Returntype  : String
+  Exceptions  : none
+  Caller      : general
+  Status      : Stable
+
+=cut
+
+sub toString {
+    my ( $self ) = @_;
+
+    # FIXME: decide what else to add here
+    return join( ', ',
+            'process=' . $self->meadow_user() . '@' . $self->meadow_host() . '#' . $self->process_id(),
+    );
+}
+
+
 1;
 
--- a/modules/Bio/EnsEMBL/Hive/Beekeeper.pm
+++ b/modules/Bio/EnsEMBL/Hive/Beekeeper.pm
@@ -202,9 +202,9 @@ sub check_if_blocked {
 sub toString {
     my ( $self ) = @_;
 
-    # FIXME: decide what else to add here
     return join( ', ',
             'process=' . $self->meadow_user() . '@' . $self->meadow_host() . '#' . $self->process_id(),
+            "options='" . $self->options() . "'",
     );
 }
 
--- a/modules/Bio/EnsEMBL/Hive/DBSQL/BeekeeperAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Hive/DBSQL/BeekeeperAdaptor.pm
@@ -150,15 +150,14 @@ sub reload_beekeeper_is_blocked {
 =cut
 
 sub block_all_alive_beekeepers {
-  my ( $self ) = @_;
+    my ( $self ) = @_;
 
-  my $statement =
-    'UPDATE beekeeper SET is_blocked = 1 WHERE cause_of_death IS NULL';
-  my $sth = $self->dbc()->prepare( $statement );
-  $sth->execute();
-  $sth->finish();
+    my $statement = 'UPDATE beekeeper SET is_blocked = 1 WHERE cause_of_death IS NULL';
+    my $sth = $self->dbc()->prepare( $statement );
+    $sth->execute();
+    $sth->finish();
 
-  return;
+    return;
 }
 
 
--- a/modules/Bio/EnsEMBL/Hive/DBSQL/BeekeeperAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Hive/DBSQL/BeekeeperAdaptor.pm
@@ -131,5 +131,35 @@ sub reload_beekeeper_is_blocked {
 }
 
 
+=head2 block_all_alive_beekeepers
+
+  Example     : $bk_adaptor->block_all_alive_beekeepers();
+  Description : Set is_blocked for all beekeepers known to the
+                pipeline which haven't died yet. Part of the "shut
+                everything down" feature - as eHive stands we cannot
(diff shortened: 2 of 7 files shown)


modules/Bio/EnsEMBL/Hive/Queen.pm
```diff
@@ -384,6 +384,46 @@ sub register_worker_death {
 }
 
 
+sub kill_all_workers {
+  my ( $self, $valley ) = @_;
```
muffato: eHive uses 4 spaces for indentation

scripts/beekeeper.pl
```diff
@@ -507,6 +514,49 @@ sub register_beekeeper {
     return $beekeeper;
 }
 
+
+sub big_red_button {
+  my ( $self, $valley ) = @_;
+
+  my $bk_a = $self->{dba}->get_Beekeeper();
```
muffato: The convention is to call `get_BeekeeperAdaptor` (the only exception is Queen, for historical reasons)

scripts/beekeeper.pl
```diff
@@ -507,6 +514,49 @@ sub register_beekeeper {
...
+  # Report which beekeepers we have just blocked
+  $blocked_beekeepers = $bk_a->fetch_all( 'is_blocked = 1' );
+  my @currently_blocked = grep {
+    ! exists $previously_blocked_ids{ $_->dbID() }
+  } @{ $blocked_beekeepers };
+  while ( my $blocked_bk = shift @currently_blocked ) {
+    # FIXME: decide what else to print
+    printf( "Blocked beekeeper: %10d %35s %35s %15s\n",
+            $blocked_bk->dbID(), $blocked_bk->meadow_host(),
+            $blocked_bk->meadow_user(), $blocked_bk->process_id() );
```
muffato: All objects but Beekeeper have a `toString` method. Maybe this is the right time to introduce one ?

modules/Bio/EnsEMBL/Hive/DBSQL/BeekeeperAdaptor.pm
```diff
@@ -131,5 +131,35 @@ sub reload_beekeeper_is_blocked {
...
+  Caller      : beekeeper.pl
+  Status      : Stable
+
+=cut
+
+sub block_all_alive_beekeepers {
+  my ( $self ) = @_;
+
+  my $statement =
+    'UPDATE beekeeper SET is_blocked = 1 WHERE cause_of_death IS NULL';
```
muffato: In eHive there is no formal maximum line length. I personally find such line breaks are breaking the flow.

modules/Bio/EnsEMBL/Hive/Queen.pm
mkszuba: Oops. Fixed.

scripts/beekeeper.pl
mkszuba: Ah, historic reasons. Fixed.

scripts/beekeeper.pl
mkszuba: Good idea! There is one now and _big_red_button()_ uses it; might make sense to make it report more, though.

Moreover, _Queen::kill_all_workers()_ now uses _Worker::toString()_ instead of crafting its own summary string.

modules/Bio/EnsEMBL/Hive/DBSQL/BeekeeperAdaptor.pm
mkszuba: Personally I like it when lines don't go too far to the right but in this case this wouldn't go _that_ far, concatenated the two.

modules/Bio/EnsEMBL/Hive/Beekeeper.pm
```diff
@@ -187,5 +187,27 @@ sub check_if_blocked {
...
+  Exceptions  : none
+  Caller      : general
+  Status      : Stable
+
+=cut
+
+sub toString {
+    my ( $self ) = @_;
+
+    # FIXME: decide what else to add here
```
ens-bwalts: I would add the options that the beekeeper was called with. Certainly the DB URL or reg_conf

scripts/beekeeper.pl
```diff
@@ -507,6 +514,47 @@ sub register_beekeeper {
...
+    $blocked_beekeepers = $bk_a->fetch_all( 'is_blocked = 1' );
+    my %previously_blocked_ids;
+    while ( my $blocked_bk = shift @{ $blocked_beekeepers } ) {
+        $previously_blocked_ids{ $blocked_bk->dbID() } = 1;
+    }
+
+    # Begin the shutdown by blocking all registered beekeepers so that
+    # none of them start spawning new workers just as this one tries to
+    # kill all workers.
+    $bk_a->block_all_alive_beekeepers();
```
muffato: This is a bit awkward because it will block the red-button beekeeper itself, even though we know it's going to exit right after

scripts/beekeeper.pl
```diff
@@ -507,6 +514,47 @@ sub register_beekeeper {
...
+
+    # Report which beekeepers we have just blocked
+    $blocked_beekeepers = $bk_a->fetch_all( 'is_blocked = 1' );
+    my @currently_blocked = grep {
```
muffato: I understand `currently_blocked` as the "beekeepers that are blocked at this current time", but what you _grep_ there is the list of say "newly blocked" beekeepers

scripts/beekeeper.pl
```diff
@@ -507,6 +514,47 @@ sub register_beekeeper {
...
+        ! exists $previously_blocked_ids{ $_->dbID() }
+    } @{ $blocked_beekeepers };
+    while ( my $blocked_bk = shift @currently_blocked ) {
+        print 'Blocked beekeeper ' . $blocked_bk->dbID() . ': '
```
muffato: Maybe just exclude the red-button bekeeper there ?

scripts/beekeeper.pl
mkszuba: Fixed.

scripts/beekeeper.pl
mkszuba: Good idea.

muffato:
Looks very good, thanks !

PR #68, opened 2019-02-08
Workaround for IPC::Run compilation failure at pipeline initialisation time

ens-bwalts:
Precompiles IPC::Run before exec is overridden.


ens-bwalts:
Closing in favour of the more generic solution in #69

PR #69, opened 2019-02-11
 Workaround for IPC::Run compilation failure at pipeline initialisation time

muffato:
This is an alternative to #68.

Instead of having to list every module that runs `exec` like this:
```
exec { $args[0] } @args;
```

I've just tried changing the scope of the override and it works on my laptop. Let's give it a try on Travis and the farm !


ens-bwalts:
Fixes the issue on the farm

PR #72, opened 2019-02-14
End of support for older Perl versions

muffato:
I've updated the docs and .travis.yml

I've also enabled testing on Perl 5.26, which we didn't test before. Fingers crossed it won't reveal any bug !


muffato:
Yeah, Travis is happy !

PR #75, opened 2019-02-14
add a pull request template

ens-bwalts:
This template is based on the one used by many of the core Ensembl repos. With the increasing number of contributors, perhaps it's time to introduce one?


muffato:
Yes it'd be useful to have one. However I don't find the order of the sections very natural (I know it's a copy of ensembl and ensembl-rest).

I would start with "Use case" (describe the problem, what is the persone trying to achieve) then "Description" (describe the solution -implementation-, how they're achieving it). I would skip "Benefits": in all the pull-requests that I've seen, it's just the same as the use case ("before we couldn't do X, but now we can"). Then keep the rest as it is.

Your introduction is good and has all the eHive specific bits I can think of.

PR #77, opened 2019-02-14
Feature/formatter

mira13:
## Use case

Formatter class added, 
The goal is consistent printing code and output, 
Example of usage is applied for Queen.pm functions called from beekeeper.pl

## Description
Formatter object can be used to set errors, warnings and info print, also it allows to apply custom output functions and set up output format - to json or text, immediate or by demand.

Import the module and use the variable

## Possible Drawbacks

None I can think of

## Testing

The test file is added, that covers 100% of formatter code

All fine

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -14,6 +14,7 @@ on 'test' => sub {
 	requires 'Test::Exception';
 	requires 'Test::More';
 	requires 'Test::Warn';
+	requires 'Test::JSON';
 	requires 'Test::Warnings';
 	requires 'Test::File::Contents';
 	requires 'Test::Perl::Critic';
@@ -25,4 +26,3 @@ recommends 'Getopt::ArgvFile';
 recommends 'BSD::Resource';
 recommends 'Chart::Gnuplot';
 recommends 'GraphViz';
-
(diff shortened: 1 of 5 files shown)


mira13:
@muffato @ens-bwalts Please review this pull request

muffato:
Can you fill the PR template to explain what this is for ?

mira13:
@muffato PR template is filled now

muffato:
Still not very clear. What problem are you trying to solve ?

The output of beekeeper has changed. It doesn't print the list of analyses any more.

mira13:
> On 7 Mar 2019, at 11:27, Matthieu Muffato <notifications@github.com> wrote:
> 
> Still not very clear. What problem are you trying to solve ?
> 
> 


Now we have many different types of output in code - direct printing, adding to hash and print late, and to warnings array,
So with formatter  and in code we have one way to print- using formatter. 
> The output of beekeeper has changed. It doesn't print the list of analyses any more.
> 
> 


This is a bug, will fix it and let you know

muffato:
But currently, the Formatter is only used as a proxy to `print`. Do you plan to use the Formatter in more places / with more formats ?

BTW, indentation is 4 spaces, not 2 (in Queen.pm)

mira13:
1.Current issue - to remove all prints from queen.pm is impossible to do without some wrapper. 
Because there is a custom print function for analysis, and if I just push all prints in one array and print after Queen.pm is finished,  either analysis will be not printed, if we comment this code or will be printed before all output.

In tweak_pipeline currently I have just commented custom output code.

2. Yes, every new format - means now new return variable, that also affects files that call the issued script.
And with formatter - only in issued script we add new code for new format, and formatter takes al care about printing.

PR #78, opened 2019-02-22
Fix lost connection regex from mysql

MatBarba:
When the mysql connection is lost, mysql clients may send 2 different
messages. This change adds the second one:
"Lost connection to MySQL server during query"


muffato:
Duplicate of #73

PR #80, opened 2019-02-27
For maintainability use the list of statuses defined in AnalysisJobAdaptor

muffato:
## Use case

I had missed this occurrence when I introduced the variable in 1d3cab12016a667263827192cc8249e56eedfb7d
The goal was to minimize the number of places with an hardcoded list of statuses so that most of the code will still work if somehow the list is changed

## Description

Import the module and use the variable

## Possible Drawbacks

None I can think of

## Testing

No, there is currently no test for `runWorker.pl --job_id`. Should I add one ? I can maybe add this to `t/03.scripts/beekeeper_reset.t`:
1. Try running a worker on the job_id of a READY job. It should return 0 and the job will be DONE
2. Try running a worker on the same job_id. It should return !=0 because the job is DONE. `when_completed` should not change
3. Try running a worker on the same job_id but with `--force`. It should accept it and `when_completed` will change

All fine

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Queen.pm
+++ b/modules/Bio/EnsEMBL/Hive/Queen.pm
@@ -77,6 +77,8 @@ use Bio::EnsEMBL::Hive::Scheduler;
 use Bio::EnsEMBL::Hive::Valley;
 use Bio::EnsEMBL::Hive::Worker;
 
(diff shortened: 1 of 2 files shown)


muffato:
Let me first extend the test before merging this

PR #82, opened 2019-02-28
Have Sphinx use sphinx.ext.imgmath instead of sphinx.ext.pngmath

mkszuba:
## Use case

The Sphinx extension _sphinx.ext.pngmath_ is deprecated, which presumably means it will disappear at some point. There seem to be only two places in eHive documentation which rely on this yet we do use it.

## Description

_sphinx.ext.imgmath_ is a drop-in replacement for it _sphinx.ext.pngmath_ (its default output format is PNG and under the bonnet both use dvipng) so all this PR does is update the Sphinx configuration file to use the former. Now not only do we no longer get the deprecation warning but could now easily switch to SVG output by passing '-D imgmath_image_format=svg' to _sphinx-build_.

## Possible Drawbacks

Sphinx versions older than 1.4 will no longer be supported. Mind you, 1.4 just under 3 years old now and our requirements file already demands 1.6.5.

## Testing

no

N/A

I have rebuilt HTML documentation with _imgmath_, for both PNG and SVG output. No problems in either case.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -50,7 +50,7 @@ from xhive import setup_if_needed
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.coverage',
-    'sphinx.ext.pngmath',
+    'sphinx.ext.imgmath',
     'sphinx.ext.graphviz',
     'sphinxcontrib.doxylink',
     'xhive.code_doc',


muffato:
Hi @mkszuba 
Have you tried this on the EBI farm too ? I remember @ens-bwalts tried this change about 2 years ago (298cea87274c3d179ac50ef8edf1a18cdccd7cb8) and for some reasons we downgraded it soon after (97db630d65685bb9ee0dc2ec8828faba936b7a39). @ens-bwalts : do you remember why ?
Apart from that, I don't mind the change. In fact I have this uncommitted change locally as I got fed up of seeing the deprecation warning :)

ens-bwalts:
I can't remember the exact issue, or whether it was with building on the farm or building at readthedocs.io. Should be tested both places, though.

mkszuba:
Looks like there is indeed a problem on the farm - LaTeX documents cannot be compiled when imgmath is used because the latter requires the LaTeX package _anyfontsize_, which is not installed.

muffato:
Is there any plan / solution to get _anyfontsize_ installed on the EBI farm ? If not, I would suggest to get the host name and select `pngmath` / `imgmath` accordingly (e.g. always the latter except on `ebi-cli-00X` machines)

mkszuba:
As far as I can tell there is no RHEL7 package containing _anyfontsize_, likely because their version of TeXLive is really old. Moreover, distro-packaged versions of TeXLive do not really support installation of third-party modules so we would most likely have to install native TeXLive, which comes with its own package manager _tlmgr_, in our software environment. Which might be tricky because not only is TeXLive pretty big, we would also have to put as much effort as possible as preventing falling through back to the system installation for certain tools or modules, which could lead to unpredictable results.

Come to think of it, there is still something I haven't tried so maybe we will be able to sidestep the above after all. I don't think I'll be able to try it before Monday afternoon, though.

mkszuba:
Found a way of getting it to work. Before launching _make html_:

`export TEXMFHOME=/tmp/my_texmf
mkdir -p $TEXMFHOME/tex/latex
wget http://mirrors.ctan.org/macros/latex/contrib/anyfontsize.zip && \
unzip anyfontsize.zip anyfontsize/anyfontsize.sty && \
rm -f anyfontsize.zip
`

I reckon we could even enable this globally, _anyfontsize_ hasn't been updated since 2007 so there is essentially zero chance of accidentally overriding a newer version.

Will think of a nice way of incorporating this into the set-up script tomorrow.

mkszuba:
Actually, there isn't any set-up script to add this to. I have just double-checked with Brandon and the only scripts we have got in the repository are for updating the RtD Docker images, and on RtD imgmath already works. This means that if anything, it is the EBI-farm environment that needs to be updated.

Moreover, I have just learned from Brandon that he has already had to install a newer version of TeXLive for himself on the farm in order to be able to build eHive documentation as PDF. That is a rather more far-reaching change than just adding a single new package for the system TeXLive to use.

All in all, I would say we should go ahead and merge this?

muffato:
Fine by me

PR #83, opened 2019-02-28
docs/xhive: support both python2 and python3

mkszuba:
## Use case

See ENSCORESW-3081: Read the Docs is now actively pushing towards using Python3 for documentation building, the custom modules we use in the process are Python2-specific, and we would like to make the switch before RtD drop Python2 support - which is likely to happen sooner rather than later because Python2.7 reaches end of life in 2020.

## Description

Adapt Python2-specific code in our custom documentation-building modules so that it works with both Python2 and Python3. This is done in a Python3-centric way in order to minimise the effort of dropping Python2 support once we have decided to do so.

## Possible Drawbacks

Slight overhead incurred by cross-version compatibility code. Temporary files are no longer opened in binary mode so there is a risk of the code no longer working as expected under Windows, assuming it has ever worked under Windows.

## Testing

No

n/a

I have successfully built HTML documentation using updated code under both Python2.7 and Python3.6. Haven't tried 3.7 but as far as I can see in the documentation, none of the changes between 3.6 and 3.7 are relevant to us.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 python:
-  version: 2.7
+  version: 3.6
   install:
     - requirements: requirements.txt
 
--- a/docs/xhive/__init__.py
+++ b/docs/xhive/__init__.py
@@ -14,6 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# For python2 compatibility
+from __future__ import print_function
+
 import json
 import os.path
 import subprocess
@@ -56,7 +59,7 @@ def setup_if_needed(this_release, run_doxygen):
             subprocess.check_call([mkdoxygen_path, doxygen_target])
 
     with open(release_holder, "w") as fh:
-        print >> fh, this_release,
+        print(this_release, end=' ', file=fh)
 
     return doxygen_target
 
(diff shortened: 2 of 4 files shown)


ens-bwalts:
@mkszuba Would you be willing to update the Python version in the newly added .readthedocs.yml and add that change to this PR. We may need to update Default version under advanced settings in the website config along with that change.

muffato:
@ens-bwalts : (I should have asked this in #84) Is it possible to unset this config option in the web config so that the .yml file is used ?

ens-bwalts:
I can't see a way to unset - the pulldown is a binary choice. I'm not 100% sure on what in the webconfig overrides the config file. There was a lot of trial and error involved in getting the config file to work, as the behaviour I was seeing was not matching the behaviour described in the documentation.

muffato:
:+1: for the rebase, I was going to ask that :)

mkszuba:
> @mkszuba Would you be willing to update the Python version in the newly added .readthedocs.yml and add that change to this PR.

Done.

ens-bwalts:
This actually needs to be Python 3.6 to use the stable build image - apologies for not catching that earlier. This is documented at https://github.com/rtfd/readthedocs-docker-images

mkszuba:
Ah, I have used the version number from the JIRA ticket. Should be fine now.

PR #86, opened 2019-03-05
Shared module to detect the SQL errors

muffato:
## Use case

This is a follow-up of #73 . Several modules check the SQL error to reconnect, restart the transaction, etc, and sometimes several modules share the exact same piece of code.

## Description

I have created a new module that catalogues the various error messages known to each driver. Consumers (usually *DBConnection and StatementHandle) can then use helper functions to detect whether the message is of a certain error or not

## Possible Drawbacks

None I can think ok

## Testing

No, there are already some tests for that

N/A

Yes


muffato:
@avullo : usually we select "rebase and merge" in the dropdown menu to linearise the commit history

avullo:
@muffato sorry I wasn't aware of your protocol, are you ok-ish anyway or do we have to roll back?

muffato:
No, it's OK. My branch was on top of master, so it makes a little bubble but it's still pretty clear

PR #91, opened 2019-04-04
[ENSCORESW-3077] Travis: Unify configuration of Slack notifications

mkszuba:
Use the same Slack access token for #coregithub as well as the same notification configuration for all Infrastructure repositories. Notifications to #ehive-commits are still sent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,10 @@ notifications:
     on_success: always
     on_failure: always
   slack:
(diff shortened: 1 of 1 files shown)


muffato:
There is already a section for Slack notifications. Will Travis report to both channels ?

mkszuba:
There were no Slack notifications configured for _version/2.2_ so I have only added a hook for #coregithub. For _version/2.3_ and newer I will have to make a separate PR which allows Travis to report to both #coregithub and #ehive-commits (which the Travis app for Slack can do), tomorrow though.

muffato:
What about https://github.com/Ensembl/ensembl-hive/blob/version/2.2/.travis.yml#L30-L31 ?

mkszuba:
Yeah, either I'm blind or I accidentally added a typo to the highlight pattern (which I had set to "notific") while looking at this repository. Should be in order now.

mkszuba:
Good idea. Done.

PR #100, opened 2019-04-24
Feature/log formatter

mira13:
## Use case

eHive needs unified output , there fore wrapper for log4perl created

## Description

Logger is used to initialise log4perl with command line parameters, 
Json and text output has different log settings.

## Possible Drawbacks

existing warnings can mixed up with new log info

## Testing

Need a discussion if any test are needed


muffato:
I think there are too many command-line parameters in `tweak_pipeline.pl` and too many combinations of outputs in Logger.
1. What is the difference between setting a logfile in log4perl and redirecting the standard output to a file ?
2. How many formats are we seriously considering at the moment ? If it's just text and json, we can keep text as the default, and use the `-json` parameter to switch to json. But I also expect that the JSON output of the scripts is likely going to be a single big structure, and the only output of the script. It's not something that fits really well into a logger, which is more about having several levels of outputs, etc.

Other comments:
 - the JSON output should be the only output on that channel (file / stdout / stderr) so that it can be parsed. That means no header at all, no timestamp, etc. I think you've addressed this via the layout and `%m` ?
 - warnings and errors, e.g. `Tweak.Error` need to be sent via the appropriate method (i.e. not always `$logger->info`)
 - get_*Logger always calls `Log::Log4perl::get_logger`. Is the same logger returned every time ? Is it not expensive to call again and again ?

About the code itself, I'm happy about:
- all the `$logger->info` calls in HivePipeline.pm (although you need to use the correct log level)
- how Utils::Logger fits

nerdstrike:
This appears to have stalled.

ens-bwalts:
Per discussion during February 2020 task planning, we are closing this PR for now (but keeping the branch). Will re-open when development on this task resumes. Comments copied to ENSCORESW-3124 to make them easier to find when this task is resumed.

PR #104, opened 2019-04-30
Java wrapper now build with maven

mira13:
## Use case

To use new versions of libraries with known licenses and modern build system, java wrapper moved to maven.

## Description

1. Maven pom.xml created
2. Folder structure changed, according with maven requirements
3. Local versions of libs removed

## Possible Drawbacks

 Class usage that is not supported any more - commented out (SimpleLogger doesn't support debug levels, should be replaced with neighbour class)

## Testing

not applicable, travis file needs fix tho, as we don't use ant any more

no

yes


muffato:
Why did you have to move all the source Java files for maven ?

.travis.yml
```diff
@@ -45,12 +45,10 @@ before_install:
     - export PATH=/usr/lib/jvm/java-8-oracle/jre/bin/:$PATH
     - export JAVA_HOME=/usr/lib/jvm/java-8-oracle
     - java -version
-    - ant -version
```
muffato: Can you replace this with ```mvn --version``` to make sure which version of maven is installed and which version of Java it is using ?

.travis.yml
```diff
@@ -45,12 +45,10 @@ before_install:
...
     - mysql --version
     - sqlite3 --version
     - env
     - cd wrappers/java
-    - ant clean jar
```
muffato: Can you make maven compile the source code here (i.e. the equivalent of `ant jar`) ? I'm sure calling `mvn test` will trigger a compilation automatically, like `ant run_tests` would, but again this is to make sure everything is ok early

wrappers/java/src/main/java/org/ensembl/hive/RunWrapper.java
```diff
@@ -48,24 +48,26 @@ public static void main(String[] args) throws Exception {
 			System.err.println("The debug level '" + args[3] + "' is not an integer. Aborting");
 			System.exit(1);
 		}
+        // NOTE: curret version of library suppose only one debug level fod SimpleLogger,
```
muffato: This block you've commented out was actually useful: depending on the selected log-level, more or less debug information would be printed, possibly some very detailed stuff ("TRACE")

wrappers/java/src/main/java/org/ensembl/hive/RunWrapper.java
```diff
@@ -23,11 +23,11 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import sun.misc.SharedSecrets;
+import jdk.internal.misc.SharedSecrets;
```
muffato: I'm quite happy that this doesn't rely on Sun stuff exclusively

wrappers/java/pom.xml
```diff
@@ -0,0 +1,59 @@
+<project>
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.ensembl.hive</groupId>
+    <artifactId>eHive</artifactId>
+    <version>1.0</version>
```
muffato: What does this version represent ? There is already a version defined in src/main/java/org/ensembl/hive/BaseRunnable.java

wrappers/java/src/main/java/org/ensembl/hive/RunWrapper.java
mira13: This library contains several classes, so we need just to use another class, not SimpleLogger, I suppose to do it in another pull request.
Here is description that Simple logger doesn't support levels https://www.slf4j.org/manual.html
, with the pic of different classes of the lib

mira13:
I have made the changes according with licensing issue, you suggest there to move to the mordent build system, that maven is
https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-2902

With maven you don't have local versions of libraries as with ant, but you always download it, it guarantee  all developers have the same libs, also it allows to see licenses easy online in maven repo, and all licenses now is compatible with Apache2, 

About standard maven folder layout - changing it - it would be like we change standard classnames convention and start class name with not-capital letter, we can do it, but it would be rather freaky.

muffato:
I :+1: the move to maven.

A comment about ENSCORESW-2902. The potential problem of embedding external jars that have a different license started in the version 2.5 when we added support for Java. I don't know if we're doing anything wrong, that's the point of that JIRA ticket. If the licensing really is an issue, the fix should be put on 2.5 and merged into master. If not, I suggest you resolve ENSCORESW-2902 itself and open another JIRA ticket about changing the build+dependency system ?

muffato:
And once you get the correct combination of `apt` packages etc, can you also update `docker/Dockerfile` ? (maybe after #105 has been merged)

wrappers/java/src/main/java/org/ensembl/hive/RunWrapper.java
muffato: I don't follow. In that page you can see they call `logger.debug`, `logger.info`. And then on https://www.slf4j.org/faq.html there is `logger.error` too

muffato:
Since I can make this work on my laptop with Java 8 and `sun.misc.SharedSecrets`, do you really need to change the Java version, the Ubuntu version etc ? If we can avoid doing unnecessary change, that'd be better.

But still, you're raising an interesting question: which JDKs (version / provider) have `sun.misc.SharedSecrets` ? And which ones have `jdk.internal.misc.SharedSecrets` ?

muffato:
You define two _repositories_ in `pom.xml`. Can you indicate which dependencies come from which repositories ?

mira13:
Regarding slf4j : I tried to put version 1.7 in maven xml , but it failed to download, that version - even is announced in the list - not any more on the repository server (file doesn't exist).

I propose just to move to the new version wit the next pull request (I mean replace commented lines with  usage new library version )

mira13:
Also repositories in pom distinguished  by name(I put name here, that shows which lib it's belonging) So now it's ready to merge.

muffato:
1. Don't forget you need to put changes on both `version/2.5` and `master`
2. Rather than disabling a feature and introducing it again soon after, directly use the correct version. But I'm very surprised that there are no 1.7.* packages available. 1.8 is beta and shouldn't be used in production
3. There is something strange with the compilation phase. Normally, one should be able to clone the repo and run `standaloneJob.pl` immediately. On 2.5 and master, this is because the jar is automatically generated. Somehow on this branch it doesn't work any more

muffato:
Forget about point n°3. This is in fact not working with `ant` either. I thought it was working because I still had the .jar in lib/. I have a fix in #108

mira13:
Now ready to merge

muffato:
I have added a commit to fix and improve `pom.xml`

`pom.xml` now only lists the direct dependencies (since _maven_ takes care of the indirect ones). This solves two issues that `pom.xml` had:
- different versions of slf4j required
- duplicated entries

For instance, it fixes this warning,
```
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.ensembl.hive:eHive:jar:1.0
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.slf4j:slf4j-api:jar -> version 1.7.5 vs 1.7.12 @ line 65, column 17
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
```

I have also changed the versions to be the same ones as in the original `build.xml`. The PR is about changing the build system, not the list / versions of the dependencies.

wrappers/java/pom.xml
```diff
@@ -0,0 +1,107 @@
+<project>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.ensembl.hive</groupId>
+  <artifactId>eHive</artifactId>
+  <version>1.0</version>
```
muffato: This should be 3.1

wrappers/java/wrapper
```diff
@@ -33,23 +33,25 @@ wrapper_dir=$(dirname "$0")
...
 	if [[ -z "$module" || -z "$fd_in" || -z "$fd_out" || -z "$debug" ]]; then
 		echo "Syntax error"
 		echo "$usage"
 		exit 1
 	fi
+
 	#echo "Running in $PWD:" java -cp "lib/*" org.ensembl.hive.RunWrapper "$module" "$fd_in" "$fd_out" "$debug"
-	exec java -cp "lib/*" org.ensembl.hive.RunWrapper "$module" "$fd_in" "$fd_out" "$debug"
+  
+	exec java -cp target/eHive-1.0-jar-with-dependencies.jar org.ensembl.hive.RunWrapper "$module" "$fd_in" "$fd_out" "$debug"
```
muffato: Use a wildcard here because the version will change

wrappers/java/wrapper
```diff
@@ -33,23 +33,25 @@ wrapper_dir=$(dirname "$0")
...
 elif [ "$action" == "compile" ]; then
 	if [[ -z "$module" ]]; then
 		echo "Syntax error"
 		echo "$usage"
 		exit 1
 	fi
-	#echo "compiling:" ant "-Dorg.ensembl.hive.module=$module" compile_wrapper
-	exec ant "-Dorg.ensembl.hive.module=$module" compile_wrapper
+	#echo "compiling:" mvn package
+	exec mvn package
```
muffato: `mvn package` is not the same as what `ant` was doing. This command should compile the wrapper _and_ run the `org.ensembl.hive.CompileWrapper` module with `$module` (the name of the Runnable that eHive wants to use

.travis.yml
```diff
@@ -42,15 +42,14 @@ before_install:
...
     - java -version
-    - ant -version
+    - mvn -version
     - mysql --version
     - sqlite3 --version
     - env
     - cd wrappers/java
-    - ant clean jar
+    - mvn clean
+    - mvn install
```
muffato: Should be `mvn package`, like in the wrapper

mira13: Here we need to run tests as well, and in wrapper only create a jar, that's why here is install.

muffato: No, tests are run in `scripts/dev/travis_run_tests.sh`. This line is to pre-compile the java code outside of the test framework, which makes it easier to debug

mira13:
Now marched against 2.5

mira13:
Will make another pull request against 2..5

PR #105, opened 2019-05-09
bugfix: oracle-java8 is not available any more, but openjdk-8 works

muffato:
## Use case

Docker builds have broken recently, because `apt` can't install the Oracle's Java version. See https://cloud.docker.com/u/ensemblorg/repository/registry-1.docker.io/ensemblorg/ensembl-hive/builds/6ade0715-5644-474d-8799-e157bdbac273

## Description

I've simply moved to openjdk. I can't remember why we needed oracle's version in the first place.

## Possible Drawbacks

If anything requires Oracle specifically, but @mira13 is also leaning towards openjdk in #104 

## Testing

I have rebuilt the Docker image locally and ran the example Java pipeline with ```prove -v /repo/ensembl-hive/t/10.pipeconfig/guest_language.t```

It worked

Yes (within the Docker image)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -35,13 +35,9 @@ FROM ubuntu:16.04
 # Install git and java (taken from https://github.com/dockerfile/java/blob/master/oracle-java8/Dockerfile)
 ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update -y \
-    && apt-get install -y git software-properties-common \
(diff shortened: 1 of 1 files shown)


muffato:
How to run the test-suite in the Docker image:
```
docker run -e EHIVE_TEST_PIPELINE_URLS=sqlite:/// ensemblorg/ensembl-hive prove -r /repo/ensembl-hive/t
```

PR #106, opened 2019-05-10
bugfix: unstored objects don't have a dbID, so this raises a warning

muffato:
## Use case

I found several warnings about performing an addition on unitialized values when running the test suite locally.

## Description

The warnings come from `apply_tweaks`. The method assumes that all analyses have a dbID, which is forced into a numeric with `+ 0`. However, `init_pipeline.pl` applies the tweaks before storing the objects, so dbID is still undefined at this stage.

## Possible Drawbacks

None

## Testing

No

Yes: OK


muffato:
Well spotted. I've now changed that too

PR #107, opened 2019-05-15
Tweak report during init_pipeline.pl

muffato:
## Use case

When initializing a pipeline with tweaks we currently see this:
```
$ init_pipeline.pl Bio::EnsEMBL::Hive::Examples::LongMult::PipeConfig::LongMultSt_javaconf $(mysql-local details hive) -tweak 'pipeline.param[take_time]=0'
(...)
> Applying tweaks.
ARRAY(0x2759b18)
> Storing the pipeline in the database.
(...)
```

1. We should be printing the actual tweaks, not a reference to a Perl array
2. The tweaks should only be displayed when `hive_debug_init` is set

## Description

The commits fix these two issues. Also, init_pipeline is now aware that the tweaks already contain `\n`, so no need to add it again.

## Possible Drawbacks

N/A

## Testing

No. The init_pipeline output is not covered by tests

Yes - OK


muffato:
Any comments ?

PR #108, opened 2019-05-15
standaloneJob for guest-languages

muffato:
## Use case

I found his bug when testing #104 
```standaloneJob.pl org.ensembl.hive.longmult.DigitFactory -language java -a_multiplier 1234 -b_multiplier 4567 -debug 1```
doesn't work unless you have manually compiled the Java code. The compilation only happens automatically when running init_pipeline.pl or runWorker.pl

## Description

I'm making standaloneJob even closer to runWorker by asking the Worker itself to compile the module (after creating a Role that links the Worker to the Analysis), like in the other parts of eHive.

## Possible Drawbacks

N/A

## Testing

I will add a test soon

### DIFF
--- a/t/05.runnabledb/java.t
+++ b/t/05.runnabledb/java.t
@@ -1,0 +1,69 @@
+#!/usr/bin/env perl
+
+# Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+# Copyright [2016-2019] EMBL-European Bioinformatics Institute
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+#      http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by muffato** [comment]
I've added a TestRunnable for Java, but note that it will have to be moved to a new location to follow maven's guidelines, as done in #104

**Comment by muffato** [comment]
There is also a reference to `ant` that will have to be changed

File: t/05.runnabledb/java.t [hunk 1]
```diff
@@ -0,0 +1,70 @@
...
+use File::Basename ();
+$ENV{'EHIVE_ROOT_DIR'} ||= File::Basename::dirname( File::Basename::dirname( File::Basename::dirname( Cwd::realpath($0) ) ) );
+
+use Test::More;
+
+use Bio::EnsEMBL::Hive::Utils::Test qw(standaloneJob);
+
+SKIP: {
+    skip "java not installed", 2 unless(`java -version 2>/dev/null`);
+    skip "ant not installed", 2 unless(`ant -version 2>/dev/null`);
```
ens-bwalts: I think this will need to be updated to mvn

muffato: Yep. Will do that

PR #109, opened 2019-05-16
Feature/java maven against2.5

mira13:
## Use case

To use new versions of libraries with known licenses and modern build system, java wrapper moved to maven.

## Description

1. Maven pom.xml created
2. Folder structure changed, according with maven requirements
3. Local versions of libs removed

## Possible Drawbacks

 Class usage that is not supported any more - commented out (SimpleLogger doesn't support debug levels, should be replaced with neighbour class)

## Testing

not applicable, travis file needs fix tho, as we don't use ant any more

no

yes


muffato:
Note: delete the `feature/java_maven` branch too once this PR is accepted

wrappers/java/wrapper
```diff
@@ -33,23 +33,27 @@ wrapper_dir=$(dirname "$0")
 cd "$wrapper_dir"
 
 if [ "$action" == "version" ]; then
-	exec ant -S -q version
+	exec mvn -S -q version
```
muffato: This fails because the `-S` option doesn't exist.
I thought that the test-suite itself would fail because of that. I need to have a look.
This is supposed to be printing the version of the wrapper (currently "3.1")

wrappers/java/wrapper
```diff
@@ -33,23 +33,27 @@ wrapper_dir=$(dirname "$0")
...
 	if [[ -z "$module" ]]; then
 		echo "Syntax error"
 		echo "$usage"
 		exit 1
 	fi
-	#echo "compiling:" ant "-Dorg.ensembl.hive.module=$module" compile_wrapper
-	exec ant "-Dorg.ensembl.hive.module=$module" compile_wrapper
+	#echo "compiling:" mvn package
+	mvn package
+  exec java -cp target/eHive-*-jar-with-dependencies.jar org.ensembl.hive.CompileWrapper "$module"
```
muffato: Use the same indentation as the rest of the block

wrappers/java/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+      <version>1.7.12</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-assembly-plugin</artifactId>
+        <configuration>
+          <archive>
```
muffato: I guess this block can be removed because `fully.qualified.MainClass` is not valid. This is supposed to be the default class that is executed when running the jar, but this is not needed anyway because the wrapper always specifies the class name

wrappers/java/wrapper
muffato: This mode is not essential. It is only there for convenience so that one can run `./wrapper version` to get the version.

I have found a solution on https://stackoverflow.com/questions/3545292/how-to-get-maven-project-version-to-the-bash-command-line
```mvn -q -Dexec.executable=echo -Dexec.args='${project.version}' --non-recursive exec:exec```

**Comment by mira13** [comment]
All changes done: Fixed indentation, version call, and exceed block in pom removed.

File: wrappers/java/wrapper [hunk 4]
```diff
@@ -33,23 +33,27 @@ wrapper_dir=$(dirname "$0")
 cd "$wrapper_dir"
 
 if [ "$action" == "version" ]; then
-	exec ant -S -q version
+	exec mvn --version
```
muffato: It's not maven's version that needs to be printed here, but the project's version. See my comment there was an example command-line

wrappers/java/wrapper
```diff
@@ -33,23 +33,27 @@ wrapper_dir=$(dirname "$0")
...
 	if [[ -z "$module" ]]; then
 		echo "Syntax error"
 		echo "$usage"
 		exit 1
 	fi
-	#echo "compiling:" ant "-Dorg.ensembl.hive.module=$module" compile_wrapper
-	exec ant "-Dorg.ensembl.hive.module=$module" compile_wrapper
+	#echo "compiling:" mvn package
+	mvn package
+  	exec java -cp target/eHive-*-jar-with-dependencies.jar org.ensembl.hive.CompileWrapper "$module"
```
muffato: Indentation should be tabulation only in this file. This line starts with some whitespace

mira13:
Now changed version to proj version, removed whitespace.

wrappers/java/wrapper
```diff
@@ -33,23 +33,27 @@ wrapper_dir=$(dirname "$0")
 cd "$wrapper_dir"
 
 if [ "$action" == "version" ]; then
-	exec ant -S -q version
+	exec mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version | grep -v '\['
```
muffato: Having a pipe defeats the purpose of `exec`, which is making the wrapper return the exit code of the command (without having to do `exit $?`).
But the other issue with this solution is that it will log the downloads too, e.g.
```
Downloading: http://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-dependency-plugin/2.1/maven-dependency-plugin-2.1.jar
Downloaded: http://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-dependency-plugin/2.1/maven-dependency-plugin-2.1.jar (104 KB at 1992.3 KB/sec)
Downloading: http://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-release-plugin/2.0/maven-release-plugin-2.0.pom
Downloaded: http://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-release-plugin/2.0/maven-release-plugin-2.0.pom (8 KB at 331.9 KB/sec)
```
This other solution hides the download / build log 
```suggestion
	exec mvn -q -Dexec.executable=echo -Dexec.args='${project.version}' --non-recursive exec:exec
```

muffato:
If it's not needed for the Java code to function, it could be removed

PR #110, opened 2019-05-18
Experimental/worker temp dir

muffato:
## Use case

This addresses issues encountered by various teams recently regarding the temporary space allocated on the EBI farm. `/tmp` is smaller than what it used to be, `/scratch` is the advised replacement of `/tmp/`, and teams sometimes need a much bigger space or a shared space. Overall, this demonstrates a need to configure the default temporary directory used by Workers.

## Description

Here I introduce a new command-line option for workers: `worker_base_temp_dir` (like `worker_log_dir`) to change what to use instead of /tmp. Command-line options can be set at the resource-class level, so the option can easily be applied to an analysis. A global default can also be set in the JSON configuration file.

I have removed the possibility of defining the temp directory name in the Runnable itself. If Runnables want to use a specific path, they would have to directly use the path they want, or instead override `worker_temp_directory` (I would advise the former). Anyway, the only Runnables that were doing that are in Compara, as far as I could see, and I'll update them. Note that this causes a breaking change to the GuestLanguage interface, which I fix here too.

I have introduced a new column in the `worker` table to track the temp directory actually used, which allows beekeeper to properly clean up leftover data from killed workers (during the garbage collection). I have also moved that function from Valley to Meadow because the meadow is known at that point and there is no need to query the Valley.

## Possible Drawbacks

External users that have their own GuestLanguage wrappers will have to update those.

## Testing

Yes

Yes

Yes

## Comments

As I will be away for the next three weeks, I won't be able to follow up with your review immediately. I still wanted to open a PR to give time to review the code and start discussing the changes. If you think it is appropriate, you can also tell the eHive reps about this branch as they may want to give it a try too.
Finally, feel free to modify the branch yourselves if you want to quickly merge this feature.

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Queen.pm
+++ b/modules/Bio/EnsEMBL/Hive/Queen.pm
@@ -572,7 +572,6 @@ sub check_for_dead_workers {    # scans the whole Valley for lost Workers (but i
 
                     if( ($worker->status ne 'SUBMITTED')                 # There is no worker_temp_directory before specialization
                     and ($worker->meadow_user eq $this_meadow_user) ) {  # if I'm actually allowed to kill the worker...
-                            $worker->set_temp_directory_name( $this_meadow->config_get('BaseTempDirectory') );
                             $this_meadow->cleanup_temp_directory( $worker );
                     }
                 }
--- a/modules/Bio/EnsEMBL/Hive/Queen.pm
+++ b/modules/Bio/EnsEMBL/Hive/Queen.pm
@@ -222,7 +222,7 @@ sub create_new_worker {
     }
 
     $worker->set_log_directory_name($hive_log_dir, $worker_log_dir);
-    $worker->set_temp_directory_name( $worker_base_temp_dir );
+    $worker->set_temp_directory_name( $worker_base_temp_dir || $meadow->config_get('BaseTempDirectory') );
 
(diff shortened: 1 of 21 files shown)


muffato:
Note that since there is a schema change, I also created the sql_schema_start tag, which will have to be moved to the right commit when this PR is rebased

muffato:
Any comments ?

modules/Bio/EnsEMBL/Hive/Queen.pm
```diff
@@ -222,6 +222,7 @@ sub create_new_worker {
     }
 
     $worker->set_log_directory_name($hive_log_dir, $worker_log_dir);
+    $worker->set_temp_directory_name( $worker_base_temp_dir || $meadow->config_get('BaseTempDirectory') );
```
ens-bwalts: If BaseTempDirectory isn't set in the config, and -worker_base_temp_dir isn't set, how does temp_directory_name get populated in the worker table?

muffato: `set_temp_directory_name` will fallback to the system's default. See https://github.com/Ensembl/ensembl-hive/blob/experimental/worker_temp_dir/modules/Bio/EnsEMBL/Hive/Worker.pm#L937-L940

PR #112, opened 2019-05-29
Travis: explicitly use Ubuntu Trusty for builds

mkszuba:
## Use case

Make sure we are not affected by the ongoing change of the default Travis build environment from Ubuntu Trusty to Ubuntu Xenial, for which we are not ready yet, by explicitly requesting the former.

## Description

It has turned out that in spite of what the Travis documentation says, perl-5.14 is *not* available under the Xenial environment on Travis. We still require compatibility with that Perl version so it is necessary for us to make sure we continue to use Trusty even after the default Travis environment has been changed to Xenial, a change that is now in progress following the end of life of Ubuntu Trusty (see https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment).

## Possible Drawbacks

Ubuntu Trusty has already reached end of life.

## Testing

No, changes affect Travis only.

N/A

No, changes affect Travis only. Travis builds for the merge branch continue to succeed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: trusty
+
 language: "perl"
 
 perl:


muffato:
The test-suite can work on Travis Xenial if our fake {{psql}} is removed. The issue is that Perl 5.14 is not available on Xenial but we are meant to be compatible (and tested) on 5.14. I don't know how to select different Perl versions depending on the Ubuntu distribution. The options I see are:
- don't specify the distribution and replace 5.14 with 5.22 so that it works on both
- set the distribution to Xenial and replace 5.14 with 5.22
- set the distribution to Trusty and keep 5.14
- stop relying on Travis' perls and add our on perls with perlbrew

mkszuba:
I would recommend against option 3, Trusty has already reached end of life after all.

Odd that perl-5.14 doesn't work under Xenial, according to Travis documentation (https://docs.travis-ci.com/user/languages/perl/) they use perlbrew rather than distro packages and we should therefore be able to continue using 5.14 without problems. Maybe we should complain about this.

muffato:
On Xenial it seems they have all the versions between 5.22 and 5.30. You can check by downloading the tar `https://storage.googleapis.com/travis-ci-language-archives/perl/binaries/ubuntu/16.04/x86_64/perl-5.30.tar.bz2`

mkszuba:
Well, having looked this up on the Internet it would appear that the documentation is wrong: https://travis-ci.community/t/failure-with-perl-5-16-5-18-5-20/2458 . Fun.

muffato:
Good, I was not going crazy :smiley_cat: 
This won't affect only eHive. cf my message on Slack#core the other day. Currently Ensembl is meant to be compatible with 5.14 and recent Perls, and we've been relying on the fact Travis has all of them

mkszuba:
We (i.e. Infra) will bring this up during the next Operations meeting.

PR #137, opened 2019-08-08
Add a "build" phase to the GuestLanguage interface

muffato:
## Use case

The GuestLanguage interface requests a compilation of the wrapper every time a worker specializes into an analysis of that language. maven still taking ~10 seconds even if the wrapper is compiled and up-to-date means that the test-suite is considerably slowed down. This model also doesn't scale as potentially hundreds of workers would all be calling maven at the same time.

## Description

In this PR I introduce a new phase of the GuestLanguage interface, "build", and a new script, {{refresh_guest_languages.pl}} to build all the wrappers. I also rename the "compile" phase to "check_exists". maven is not needed at runtime any more.

This allows some drastic changes to the Docker image. The Dockerfile is now a multi-stage build, maven being pulled in just 1 phase to build the wrapper. Then the final image only needs to include the JRE, not the JDK+maven. In fact there are now several more stages in the Dockerfile. This simplifies the handling of temporary tools that are only needed for one operation but not in the final image. I have also made the Dockerfile pull the whole checkout as "build context" rather than fetching it from Github (will have to reconfigure this on the Docker Hub). It greatly simplifies testing things locally and reduces the amount of data being transferred. That plus a few other changes, the eHive Docker image is now slightly more than 700MB (about a CD-ROM ^^). It was 1.12GB before

## Possible Drawbacks

The pipeline won' be able to automatically pick up changes in the wrapper code. One would have to run {{refresh_guest_languages.pl}}. I think this is acceptable and quite typical (like having to rerun `make` in order to update the executable).

## Testing

Already covered by tests.

Yes


scripts/refresh_guest_languages.pl
```diff
@@ -0,0 +1,193 @@
+#!/usr/bin/env perl
```
ens-bwalts: I'm on the fence on whether a call to this should be added to guest_language.t, or a seperate test put under 03.scripts/. It wouldn't test much other than to be sure that build_wrapper_for_language isn't irreparably broken - so it might not be worth the file manipulation involved.

Otherwise, this looks good and congrats on getting the image down to CD size.

muffato: I can achieve in a similar way as `t/04.meadow/lsf.t`.
I create and register a wrapper for a "dummy" language. The wrapper is in fact an `ok()` in disguise, a bit like https://github.com/Ensembl/ensembl-hive/blob/version/2.5/t/04.meadow/fake_bin/bacct#L114-L123 but I would make it return 0/1 upon a certain condition to test both outcomes.
This test would test the interaction of `Bio::EnsEMBL::Hive::GuestProcess` with the wrappers:
- does it call the wrapper with the correct arguments
- I can extend it to test that it detects the wrappers correctly

But then I wouldn't have to test scripts/refresh_guest_languages.pl ?

ens-bwalts: Yeah, that's a pretty sensible way to do it.

muffato: I have added a unit-test for GuestProcess. Let me know if you think it covers enough of the wrapper interaction (incl. the methods `refresh_guest_languages.pl` calls)

PR #140, opened 2019-08-21
Throw if the requested registry file doesn't exist

muffato:
## Use case

When a worker or beekeeper is invoked with a wrong reg_conf argument, there is no explicit warning / error message, only potentially an error about the reg_alias not being found in the Registry.

## Description

`Registry::load_all` already has a flag to require the file to exist. I merely enable it and we now get a much clearer error message.

## Possible Drawbacks

This is a breaking change for people who use a url _and_ an invalid reg_conf at the same time. Previously, there would have been no warnings about the invalid reg_conf, and the database would have been connected to via its URL. Now eHive is going to complain about the reg_conf,

## Testing

Yes, but requires Ensembl/ensembl#408

Yes

Yes

### DIFF
--- a/t/01.utils/go_figure_dbc.t
+++ b/t/01.utils/go_figure_dbc.t
@@ -20,6 +20,7 @@ use strict;
 use warnings;
 
 use Test::More;
+use Test::Exception;
 use Data::Dumper;
 
 use Bio::EnsEMBL::Hive::Utils qw(go_figure_dbc);
@@ -51,5 +52,9 @@ my $dba = Bio::EnsEMBL::Hive::DBSQL::DBAdaptor->new(-url => $pipeline_url,
 						   -no_sql_schema_version_check => 1);
 my $dba2dbc = go_figure_dbc($dba);
 isa_ok($dba2dbc, 'Bio::EnsEMBL::Hive::DBSQL::DBConnection');
+
+throws_ok {
+    Bio::EnsEMBL::Hive::DBSQL::DBAdaptor->new(-reg_conf => '/non_existent_file');
+} qr/Configuration file .* does not exist. Registry configuration not loaded/, 'Throws a relevant message if the path doesn\'t exist';
 
 done_testing();
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,8 @@ before_install:
     - eval "$(perl -Mlocal::lib=${HOME}/deps)"
     - mkdir deps
     - cd deps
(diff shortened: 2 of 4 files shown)


t/01.utils/go_figure_dbc.t
```diff
@@ -52,4 +53,8 @@ my $dba = Bio::EnsEMBL::Hive::DBSQL::DBAdaptor->new(-url => $pipeline_url,
 my $dba2dbc = go_figure_dbc($dba);
 isa_ok($dba2dbc, 'Bio::EnsEMBL::Hive::DBSQL::DBConnection');
 
+throws_ok {
+    Bio::EnsEMBL::Hive::DBSQL::DBAdaptor->new(-reg_conf => '/non_existent_file');
+} qr/Configuration file .* does not exist. Registry configuration not loaded/, 'Throws a relevant message if the path doesn\'t exist';
```
ens-bwalts: Doesn't seem to match the actual error message - see failure on Travis for this test

muffato: Yes, because it needs Ensembl/ensembl#408

muffato:
One thing that's missing, though, is that the workers would only exit, without letting the beekeeper know, resulting in the beekeeper endlessly submitting new workers. See ENSCOMPARASW-731

In some cases, when both a URL and a registry file are provided, we could make the worker buffer the registry file error, connect with the URL, record the error message and then quit. Beekeeper should then find those errors and exclude the analysis (like Runnable compilation errors) ?
In other cases, when the eHive database is given as a registry alias, the worker can't possibly write anything into the database, so beekeeper would have to figure out that it's submitting workers and they never connect back. A solution for this case might actually work for the previous case too.

t/01.utils/go_figure_dbc.t
ens-bwalts: Actually, the real issue with the test is that we don't clone ensembl into Travis, therefore the `require Bio::EnsEMBL::Registry` fails.

muffato: :man_facepalming:
I hope I can get away with just cloning ensembl, without having to install many other dependencies

muffato:
FYI I'm off for a week. Will do that when I'm back

muffato:
FYI I have modified the base of the PR to demonstrate how the PR would look like once #142 has been accepted and merged into master.

Do not merge this PR until then !

muffato:
The diff will be much clearer (and there won't be any more conflicts) once #145 is approved

PR #141, opened 2019-09-12
bugfix: allow generating graphs of pipelines with a dot in their name

muffato:
## Use case

@JAlvarezJarreta complained that guiHive was unable to display his pipeline. We found that this was caused by the database name having a dot in it. `generate_graph.pl` also fails.

He found the issue on master. Here is the fix for version 2.4, which I will have to merge and adapt to the later branches, but there *will* be some conflicts as this part of the code has changed.

## Description

My fix is to quote the _subgraph_ names and remove dots from internal node names (I couldn't find how to quote these, and anyway this is only the internal name - the label works fine).

## Possible Drawbacks

None

## Testing

No. There are no tests for generate_graph.pl on 2.4 . These were added on 2.5 only.
I seem to recall that the 2.4 graph generation cannot be reliably configured from the test suite and the host configuration can "leak" into the output of graphviz, making the tests a bit random.

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Utils/Graph.pm
+++ b/modules/Bio/EnsEMBL/Hive/Utils/Graph.pm
@@ -329,6 +329,7 @@ sub _add_pipeline_label {
     my $node_fontname       = $self->config_get('Node', 'Details', 'Font');
     my $pipeline_label      = $pipeline->display_name;
     my $pipelabel_node_name = 'pipelabel_'.$pipeline->hive_pipeline_name;
+       $pipelabel_node_name =~ s/\./_/g;
 
     $self->graph()->add_node( $pipelabel_node_name,
         shape     => 'plaintext',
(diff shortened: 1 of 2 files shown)


muffato:
By the way, the bug in theory affects the previous versions, but in 2.1-2.3 there is another bug that prevents dealing with databases with a dot in the name in the first place.
That other bug is in the URL parser. None of the eHive commands (incl. `db_cmd.pl`) allows connecting to such databases by URL, which is what `init_pipeline.pl` needs to create the database. We could imagine someone first creating a database without a dot in the name and then renaming the database it, and trying to generate the graph by using a reg_conf. Bam they would see the GraphViz bug ! I think it's not worth fixing. The URL parser would be a more important thing to fix, but that's one of the reasons why it was changed in 2.4, so I'd leave it as it is as well. Maybe I could introduce a "Known bugs" section to those branches, though ?

muffato:
FYI the fix on `version/2.5` and `master` will be db04ca1d5f78e1fa6f0e814103f810245ff6f663

PR #148, opened 2019-11-20
Misc. changes re. PostgreSQL

muffato:
## Use case

There has been some transient Travis failures recently, and they seem to arise from `t/03.scripts/beekeeper_reset.t`. The problem is that runWorker doesn't always run the jobs in the same order, so the job statuses don't match what the test expects.
See https://travis-ci.org/Ensembl/ensembl-hive/jobs/614039653 and https://travis-ci.org/Ensembl/ensembl-hive/jobs/613529841

## Description

Instead of running a single generic runWorker command, run specific jobs in a certain order.

At the same time, I simplify the AnalysisJob adaptor a little bit, as we don't need to always CAST the job status in PostreSQL. It is only needed for CASE expressions, as they otherwise default to a being a string.

## Possible Drawbacks

None

## Testing

Yes

Yes

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
@@ -87,8 +87,10 @@ sub default_overflow_limit {
 =head2 job_status_cast
 
   Example     : $job_adaptor->job_status_cast();
-  Description : Returns a job-status expression that the SQL driver understands.
-                This is needed for PostgreSQL
+  Description : Changes the type of the expression to the type of the
+                job-status ENUM.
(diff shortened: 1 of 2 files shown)


ens-bwalts:
> Can the status values of READY, SEMAPHORED etc be moved to constants and then used through the constants instead of the literals?

I think that having a method to wrap the job status in a CAST statement is more efficient than having a set of constants for each (R)DBMS that eHive supports. I think it's unlikely that we'll support another DB anytime soon, but if we eventually do then we can just add whatever syntax change is needed into job_status_cast rather than have to create more constants. As it is, the actual values in the ENUM are the same for each RDBMS.

Having constants could help with maintainability, might be something to think about when there's a future schema change of the jobs table

PR #149, opened 2019-12-05
Misc. changes re. testing the text outputs

muffato:
## Use case

My motivation was to make the output of `prove -r t` clearer by removing spurious eHive messages. 

## Description

I'm achieving that by capturing stdout/stderr accordingly and checking that they are as expected. I also check that eHive issues warnings as expected (e.g. URL parsing)

## Possible Drawbacks

When changing some error messages we'll have to change the unit-test as well.
There are still some tests that will have stuff printed to stderr and that I haven't been able to silence / test yet.

## Testing

Yes

Yes

Yes

### DIFF
--- a/t/03.scripts/beekeeper_reset.t
+++ b/t/03.scripts/beekeeper_reset.t
@@ -96,14 +96,19 @@ foreach my $pipeline_url (@$ehive_test_pipeline_urls) {
     my $completion_date1 = fetch_job_completion_date($job_adaptor, $job_id);
 
     # Try again
-    ok(system($ENV{'EHIVE_ROOT_DIR'}.'/scripts/runWorker.pl', '--job_id', $job_id), 'Cannot rerun a job that is already done without the --force option');
+    #  Guarantee there would be at least 1 second between the completion timestamps
+    #  You can't trust time on Travis, sleep(1) may sleep for less than 1 second, so sleeping for 2 seconds
+    sleep(2);
+    # Won't do anything because the job is already DONE, but runWorker still returns 0
+    runWorker($pipeline_url, ['--job_id' => $job_id]);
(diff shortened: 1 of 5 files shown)


muffato:
I have removed one commit because the capture_* methods are interfering with Perl's TAP parser. I've got an idea on how to test SystemCmd's standard output / error but will do that separately.
The test-suite should now pass

t/03.scripts/beekeeper_reset.t
```diff
@@ -96,7 +96,7 @@ foreach my $pipeline_url (@$ehive_test_pipeline_urls) {
     my $completion_date1 = fetch_job_completion_date($job_adaptor, $job_id);
 
     # Try again
-    ok(system($ENV{'EHIVE_ROOT_DIR'}.'/scripts/runWorker.pl', '--job_id', $job_id), 'Cannot rerun a job that is already done without the --force option');
+    runWorker($pipeline_url, ['--job_id' => $job_id, '--force']);
```
ens-bwalts: Just to make sure my understanding is correct - what I think is going on is this: this line and the following assert_jobs weren't really doing much, beyond verifying that runWorker would respect --job_id. (it was trying to get runWorker to complain that it couldn't re-run a done job without --force, but there was nothing to observe because runWorker exits with 0.) Job 4 would still be done, so assert_jobs would show that no other jobs changed status. With your change, you're showing that you can try and force the re-run of a done job, so the 0 exit code now makes sense, and then go on to verify that no jobs changed status. Is that correct?

muffato: Wait a sec. I think I've missed that just below there is another `runWorker` call on the same job_id with `--force` too. I think this line shouldn't have `--force`, and we check that the completion date remains the same. Then the next call will use `--force` and we can check that the completion date has changed

PR #150, opened 2019-12-05
Silence codecov

muffato:
## Use case

I personally find the codecov messages too invasive. Unlike coveralls, they cannot be controlled from the codecov website, it has to be set in `codecov.yml`
This is not my first attempt. I hope this one will work ! We'll quickly see on this PR

## Description

I change `codecov.yml` to disable the comments. codecov should still notify if the coverage drops by more than 1%. I could test this by doing a fake PR that removes most of the test suite.

## Possible Drawbacks

Code coverage stats will be less visible

## Testing

This PR will test things live

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,6 +3,8 @@ coverage:
   status:
     project:
       default:
+        # Don't report failures on pull-requests
+        informational: true
   notify:
     slack:
       default:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,10 +3,8 @@ coverage:
   status:
     project:
       default:
-        threshold: 1%
   notify:
     slack:
       default:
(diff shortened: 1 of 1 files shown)


muffato:
I've done some tests on my fork. I still can't find a way to tell codecov to allow <1% coverage decrease, but I've managed to test a few other things:
 - codecov can be made "informational" only, i.e. it will report the coverage change, but will always succeed regardless of the actual coverage change
 - the PR comment can be turned off

What would you prefer ? I'd be happy to do both. I like the codecov website, I find it way better than coveralls, but the configuration it too annoying, and I don't want to get it in the way of our PRs.

Also note that in theory, codecov should be using the yml configuration file of this branch, but in my tests, it's using the one of the base branch (or the default branch, not sure). So once we've decided what to do with codecov, I may move this PR over to 2.4 and do a cascade merge to have the change applied to all branches ?

ens-bwalts:
I'm in favor of making it information only. Having the "details" link in the PR is convenient, so a way to keep that while getting rid of the some checks were not successful message is a godd thing. A bit strange that it goes for either default or master branch config, will keep that in mind for the future.

PR #155, opened 2020-04-24
Better detection of the process being killed

muffato:
## Use case

I've found cases where the process was being killed by LSF (MEMLIMIT) but there was also an exit code in the upper 8 bits. In those cases, we still want to go through the MEMLIMIT path.

## Description

I changed the condition from "not having data in the upper 8 bits) to "having data in lower 8 bits".
Note that because this module has frequently changed, it won't be a straightforward merge. The equivalent commits are:
- [version 2.4](https://github.com/Ensembl/ensembl-hive/commit/390909d31c7821f0d08e6016990a4778b983a83a)
- [version 2.5](https://github.com/Ensembl/ensembl-hive/commit/f86a618117276227c183a5298a3251ffa54c7530)

## Possible Drawbacks

None envisaged

## Testing

Not sure how to test this. Most of the time a kill results in only the lower 8 bits having some data

Yes. OK

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/RunnableDB/SystemCmd.pm
+++ b/modules/Bio/EnsEMBL/Hive/RunnableDB/SystemCmd.pm
@@ -128,15 +128,17 @@ sub write_output {
     my $stderr = $self->param('stderr');
     my $flat_cmd = $self->param('flat_cmd');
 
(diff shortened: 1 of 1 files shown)


muffato:
I have reworked the comments on 2.3. Will update the fixes for the other branches once you're happy with it

PR #156, opened 2020-04-24
Improved big red button

muffato:
## Use case

Found two issues when calling beekeeper with `-big_red_button`.
1. beekeeper tried to kill workers of another user
2. beekeeper didn't clean up the temporary directories

## Description

Both issues are quite straightforward to fix

## Possible Drawbacks

None envisaged

## Testing

1. Can't have more than 1 user in the test-suite.
2. About the temp directory, I would have to change the `longrunning` analysis of the example pipeline `LongWorker` to make it write somewhere. Can I do that ? \
By the way, this example pipeline is only referenced in the test-suite (introduced in b98c8b891df5d962934fc4b711b893170ff56727) and I don't think it has any pedagogical value (unlike `CompressFiles`, `MemlimitTest`, etc). Would you mind if I move it to `t/10.pipeconfig/TestPipeConfig` and change the analysis ?

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Queen.pm
+++ b/modules/Bio/EnsEMBL/Hive/Queen.pm
@@ -421,6 +421,9 @@ sub kill_all_workers {
                 }
             }
         }
+        else {
+            $kill_status = 'could not kill, running under user ' . $worker->meadow_user;
+        }
 
         print 'Killing worker ' . $worker->dbID() . ': '
             . $worker->toString( 1 ) . " -> $kill_status \n";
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Hive/Queen.pm
```diff
@@ -415,11 +416,14 @@ sub kill_all_workers {
                 $kill_status = 'requested successfully';
                 $worker->cause_of_death( 'KILLED_BY_USER' );
                 $self->register_worker_death( $worker );
+                if ( $worker->status ne 'SUBMITTED' ) {             # There is no worker_temp_directory before specialization
+                    $meadow->cleanup_temp_directory( $worker );
+                }
             }
         }
```
ens-bwalts, line 423: Suggest adding
`else { $kill_status = 'could not kill, running under user ' . $worker->meadow_user }`
to provide a more complete message and avoid having `$kill_status` be undef

muffato: Good point. I have fixed this in a new commit

PR #157, opened 2020-05-05
Features/python packaging

marcoooo:
## Use case

All eHive python wrapper import from pip with 
pip install -e git+https://github.com/Ensembl/ensembl-hive.git@features/python-packaging#egg=ensembl-hive

## Description

Allow setup of a dependency to Python Wrapper from a requirements.txt

## Possible Drawbacks

None as far as I can see.

## Testing

_Have you added/modified unit tests to test the changes? No

_If so, do the tests pass/fail?_ N/A

_Have you run the entire test suite and no regression was detected?_ No


ens-bwalts:
One small thing, the license check test is failing.
Need to either add licenses or modify the filename patterns near the end of `t/apache2.t` so that it will ignore the new files being added.

marcoooo:
> One small thing, the license check test is failing.
> Need to either add licenses or modify the filename patterns near the end of `t/apache2.t` so that it will ignore the new files being added.

I used the new way of adding LICENSE, sorry.

muffato:
Not sure how that's going to be used / what this is solving. You can't get a Runnable to work just from installing the wrapper like that, you need to have all of eHive installed. And once installed, you don't need to do anything to use the Python wrapper (no need to set `PYTHONPATH`), it works out of the box.

The only benefit I see is that it will put eHive in `PYTHONPATH`, so you can now import eHive outside of the runtime environment. But the cost (1 command to run once) is the same as setting `PYTHONPATH` in the bashrc / module.

Also note there is a caveat with this method: it only allows 1 version of the wrapper installed at a time. Some users may need to use different versions for different pipelines.

marcoooo:
> Not sure how that's going to be used / what this is solving. You can't get a Runnable to work just from installing the wrapper like that, you need to have all of eHive installed. And once installed, you don't need to do anything to use the Python wrapper (no need to set `PYTHONPATH`), it works out of the box.
> 
> The only benefit I see is that it will put eHive in `PYTHONPATH`, so you can now import eHive outside of the runtime environment. But the cost (1 command to run once) is the same as setting `PYTHONPATH` in the bashrc / module.
> 
> Also note there is a caveat with this method: it only allows 1 version of the wrapper installed at a time. Some users may need to use different versions for different pipelines.

Hi Matthieu, so far, the main use case is to be able to reuse the BaseRunner class locally without having to add manually the ensembl-hive path in you PYTHONPATH (If you have another solution, I am really keen to know how to do it). Moreover for testing your pipeline runner (see https://github.com/Ensembl/ols-ensembl-loader/blob/da1e54c22259b17699d3befbd5b41893544ca04e/tests/test_basic.py#L298) without having to checkout the all ensembl-hive repository in travis. Is it enough to worth your approval ;-) ?

People running multiple version of the Python wrappers shouldn't be too numerous, right? and it could be managed with different runs with proper virtualenv and an installed version from the required branch.

muffato:
I need to check this. It should be possible to run eHive Runnables (Perl & Python) *without* having to set up `PERL5LIB` and `PYTHONPATH`. If it's not the case, it's a bug that should be fixed

muffato:
Also, I would ask not to add this blanket `.gitignore`. There are already per-topic `.gitignore`, i.e.
```
$ find -name .gitignore -exec head \{\} \+
==> ./wrappers/python3/.gitignore <==
__pycache__/

==> ./wrappers/java/.gitignore <==
/build/
/*.jar
/target/

==> ./docs/.gitignore <==
_build
*.pyc

==> ./t/.gitignore <==
03.scripts/visualize_jobs/*/*.png
```

No problem adding other patterns, but certainly not this things like `[Sc]ripts` ! That's where beekeeper, runWorker, etc, are !

marcoooo:
> Also, I would ask not to add this blanket `.gitignore`. There are already per-topic `.gitignore`, i.e.
> 
> ```
> $ find -name .gitignore -exec head \{\} \+
> ==> ./wrappers/python3/.gitignore <==
> __pycache__/
> 
> ==> ./wrappers/java/.gitignore <==
> /build/
> /*.jar
> /target/
> 
> ==> ./docs/.gitignore <==
> _build
> *.pyc
> 
> ==> ./t/.gitignore <==
> 03.scripts/visualize_jobs/*/*.png
> ```
> 
> No problem adding other patterns, but certainly not this things like `[Sc]ripts` ! That's where beekeeper, runWorker, etc, are !

Sorry, I shouldn't have use a generic gitignore generator for python / perl / jetbrains projects. 
Tell me what to do

marcoooo:
> Also, I would ask not to add this blanket `.gitignore`. There are already per-topic `.gitignore`, i.e.
> 
> ```
> $ find -name .gitignore -exec head \{\} \+
> ==> ./wrappers/python3/.gitignore <==
> __pycache__/
> 
> ==> ./wrappers/java/.gitignore <==
> /build/
> /*.jar
> /target/
> 
> ==> ./docs/.gitignore <==
> _build
> *.pyc
> 
> ==> ./t/.gitignore <==
> 03.scripts/visualize_jobs/*/*.png
> ```
> 
> No problem adding other patterns, but certainly not this things like `[Sc]ripts` ! That's where beekeeper, runWorker, etc, are !

I can't compare the .gitignore file to any other version from master, where is located the one I should add line into ?

marcoooo:
> I need to check this. It should be possible to run eHive Runnables (Perl & Python) _without_ having to set up `PERL5LIB` and `PYTHONPATH`. If it's not the case, it's a bug that should be fixed

I still wonder how it's possible to be able to develop an sub class of Runnable without having BaseRunnable in the path, you'll have to show me. I can't have the class resolved in PyCharm without having set the required path in pythonpath. btw, this imply to have the whole repo checked out)

muffato:
I was only talking about runtime (haven't checked if it works, though). During dev, if you want access to the the Perl / Python modules, you indeed need to update `PERL5LIB` and `PYTHONPATH`

marcoooo:
> I was only talking about runtime (haven't checked if it works, though). During dev, if you want access to the the Perl / Python modules, you indeed need to update `PERL5LIB` and `PYTHONPATH`

Ok, so I didn't miss anything, so this is typically the use cases I created those files for then :-) To be able to develop and build with simple requirements instead of having to set all up for example for travis build. I will update the gitignore.

muffato:
Confirmed, things can _run_ without having to set `PYTHONPATH` for eHive (only for your own runnables). I moved the TestRunnable to another directory and could run
```
env PYTHONPATH=$PWD standaloneJob.pl TestRunnable -debug 9 -language python3
```

muffato:
> I can't compare the .gitignore file to any other version from master, where is located the one I should add line into ?

All four `.gitignore` are there on both master and v2.5. On v2.3 and v2.4 there was a single `.gitignore` at the root of the repo

setup.py
```diff
@@ -0,0 +1,46 @@
...
+# limitations under the License.
+
+from pathlib import Path
+
+from setuptools import setup, find_packages
+
+with open(Path(__file__).parent / 'LICENSE') as f:
+    license_ct = f.read()
+with open(Path(__file__).parent / 'wrappers' / 'python3' / 'VERSION') as f:
+    version = f.read()
```
muffato, line 23: To get the version do this instead:
```python
import eHive
version = eHive.__version__
```
No need to add the `VERSION` file (and those going out of sync)

marcoooo: That was the idea I had first. but didn't work when running setup upon pip install. I'll try again today.

marcoooo: I mean eHive is not in "PATH" when unpacking compressed built library.

marcoooo:
> > I can't compare the .gitignore file to any other version from master, where is located the one I should add line into ?
> 
> All four `.gitignore` are there on both master and v2.5. On v2.3 and v2.4 there was a single `.gitignore` at the root of the repo

... still https://github.com/Ensembl/ensembl-hive/tree/master. I can't see any .gitignore at repo root. right ?

setup.py
muffato: Right, so
```python
from wrappers.python3 import eHive
```
should work, but it doesn't yet. I will add another commit to the PR to address this and the version there

muffato:
> ... still https://github.com/Ensembl/ensembl-hive/tree/master. I can't see any .gitignore at repo root. right ?

Yes, that's correct

setup.py
```diff
@@ -0,0 +1,46 @@
...
+setup(
+    name='ensembl-hive',
+    package_dir={"": "wrappers/python3/"},
+    packages=find_packages(where='wrappers/python3/'),
+    description="Ensembl Python Base Hive Wrapper",
+    author='Ensembl',
+    author_email='dev@ensembl.org',
+    url='https://www.ensembl.org',
+    download_url='https://github.com/Ensembl/ensembl-py',
+    license=license_ct,
```
muffato, line 34: When reading about `MANIFEST.in`, I've found that there is a [`license_file`](https://setuptools.readthedocs.io/en/latest/setuptools.html#metadata) option. I think we should be using it (and in ensembl-py / ensembl-compara too)

muffato: That would mean we don't need to do `include LICENSE` in `MANIFEST.in`

muffato: And since `MANIFEST.in` would only have `include wrappers/python3/README.md`, I wonder if we could find another way of automatically bringing it in via https://packaging.python.org/guides/using-manifest-in/#how-files-are-included-in-an-sdist

marcoooo: Yes may be, please could we have a call a some point to discuss all this with our actual voices :-)

muffato: According to https://packaging.python.org/guides/distributing-packages-using-setuptools/#license `license` shouldn't be the content of the license but the title of the license with the version (and if we don't care about the version, the classifier below would be enough)

muffato: @JAlvarezJarreta . We should probably do the same in ensembl-compara, see https://github.com/Ensembl/ensembl-hive/blob/features/python-packaging/setup.py#L31

.gitignore
```diff
@@ -0,0 +1,3 @@
+.idea/**
+.eggs/**
+dist/
```
muffato, line 3: Why not adding `build/` as well ?

.gitignore
```diff
@@ -0,0 +1,3 @@
+.idea/**
```
muffato, line 1: That's an IDE-specific rule. Why only this IDE though and not all the other ones / text editors ? I would say either all or none, my preferred option being none. @ens-bwalts

.gitignore
```diff
@@ -0,0 +1,3 @@
...
+.eggs/**
```
muffato, line 2: How to get something installed there ? It's not created on my system. If it's not used, it shouldn't be listed

.gitignore
marcoooo: I can add it, no problem for me. dist is the default output dir when using setup.py, I think, but I am not sure.

.gitignore
marcoooo: The first commit I did with the "top level" git ignore did the ignore for a bunch of IDE, if I reckon well. But as well, I can remove it. really up to you guys, you are the maintainer. Also, having "all of them" may be overkill, if only we add the ones used across ensembl, that would be great :-)

.gitignore
marcoooo: did you run the setup.py sdist command ?

.gitignore
muffato: I got `build/` when running `python setup.py build` (before `python setup.py install`)

.gitignore
muffato: Yes, it only created `dist/`

marcoooo: ok so this .eggs migth be the result of a first misconfiguration of my setup.py. 
I will update accordingly.

marcoooo: Just ran a setup.py  bdist_egg. just `dist` directory created. So updated in last commit.

marcoooo:
Hi @muffato, @ens-bwalts  What's the policy here? Is it me or you who actually do the merge once PR is accepted?

ens-bwalts:
Hi @marcoooo it needs two approvals. I will take a look at this, and I'll approve and merge (unless spotting any last minute thing). Sorry, there has been back and forth on this a lot over the past few weeks and I hadn't noticed @muffato 's approval.

PR #158, opened 2020-05-13
Moved the pipeline to the test-suite since it is only useful there

muffato:
## Use case

Follow-up of #156 . I pressed _Merge_ too quickly and forgot to add the unit test.

## Description

I have moved the `LongWorker` example pipeline to the test-suite because there is nothing to learn from it. I have expanded the "big red button" test to test the deletion of the temp directory (I think that Meadow method was not yet tested, so the test coverage should slightly increase :crossed_fingers: )

## Possible Drawbacks

Anyone in deed of this PipeConfig won't be able to access it any more

## Testing

Yes

Yes

Yes

### DIFF
--- a/t/03.scripts/TestPipeConfig/LongWorker_conf.pm
+++ b/t/03.scripts/TestPipeConfig/LongWorker_conf.pm
@@ -2,11 +2,11 @@
 
 =head1 NAME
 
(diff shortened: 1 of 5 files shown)


muffato:
The tests pass locally but Travis is (genuinely) failing. I need to investigate

muffato:
The build is now passing. Cleaning up temp directories is optional, and I had to turn the option on

PR #160, opened 2020-06-02
Python bugfixes

muffato:
## Use case

Two bugs I have found when working on the Python codebase. The wrapper has changed a little bit, and I believe there will be a conflict when merging into 2.5, meaning this has to be done manually and bypass the branch protection

## Description

1. The `autoflow` attribute was used on the wrong object
2. Wrong number of arguments when calling `usage`

## Possible Drawbacks

NA

## Testing

No test updated / written. I can add some if you want. For the dataflow thing, I would have to write a new Python runnable that dataflows its own stuff onto branch #1. The wrapper has no unit-test yet


muffato:
Ah yes sorry it's done but I haven't made a PR yet.

PR #161, opened 2020-06-05
Account for time slippage on Travis

muffato:
## Use case

When running on Travis, `sleep(10)` may sleep for less than 10 seconds.
We have no explanation for this, so when it happens we simply restart the
job.

```
#   Failed test 'The "take_time" parameter made the runnable sleep a bit'
#   at t/05.runnabledb/dummy.t line 49.
#     '9.66858005523682'
#         >=
#     '10.0015641689301'
# Looks like you failed 1 test of 12.
```

## Description

`dummy.t` detects it is running on Travis and expects that 9 seconds have
elapsed instead of 10. I think all the cases we have seen are > 9.5 seconds
so we should avoid most failures.

Note that I have also removed the estimation of the overhead. I (now) think
it is not needed in this Runnable. All that matters is that we slept
enough.

## Possible Drawbacks

None

## Testing

Yes

Yes

Yes

### DIFF
--- a/t/05.runnabledb/dummy.t
+++ b/t/05.runnabledb/dummy.t
@@ -24,24 +24,13 @@ use Time::HiRes qw(time);
 
 use Bio::EnsEMBL::Hive::Utils::Test qw(standaloneJob);
 
-my $min_overhead;
-
-for (1..10) {
-    my $t = time();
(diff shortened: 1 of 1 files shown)


ens-bwalts:
Not sure why Travis didn't run on this PR. Ran the test suite myself and it passes.

PR #162, opened 2020-06-08
In standaloneJob mode there is no need to query all meadows

muffato:
## Use case

If you add the SLURM meadow to your PERL5LIB, you will notice some of its
debug statements when running stansaloneJob. It's because the constructor
of Valley calls `loaded_meadow_drivers` which lists and loads all the
available Meadow drivers.
I believe that's not necessary. In standaloneJob job the worker is pretty
much a fake one. Its purpose is to mimic the "real" execution of the job
(alongside a fake Role, a fake Analysis, a fake Job, etc) but that's all.
The Worker is not recorded anywhere, and the attributes that are pulled
from the Valley are not used anywhwere.

Anyway, I think standaloneJob shouldn't bother about all the meadows. It
represents a debugging task and as such should just assume to be running on
LOCAL.

## Description

The change sets the list of loaded drivers to LOCAL only, so that it is the
only Meadow considered by the Valley.

## Possible Drawbacks

None.

## Testing

No. The only way I could think of is to write a Meadow driver that doesn't
compile, add it to PERL5LIB, run standaloneJob, and expect it _not to
fail_. It feels a bit too much, what do you think ?

Fine

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Scripts/StandaloneJob.pm
+++ b/modules/Bio/EnsEMBL/Hive/Scripts/StandaloneJob.pm
@@ -26,13 +26,18 @@ use warnings;
 use Bio::EnsEMBL::Hive::AnalysisJob;
 use Bio::EnsEMBL::Hive::GuestProcess;
 use Bio::EnsEMBL::Hive::HivePipeline;
+use Bio::EnsEMBL::Hive::Meadow::LOCAL;
 use Bio::EnsEMBL::Hive::Queen;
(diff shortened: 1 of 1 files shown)


muffato:
I didn't think of that functionality. What would it give over 
```
bsub "description here" standaloneJob.pl My::Runnable
```
? Is it for meadows where the submission is complex, e.g. DockerSwarm ?

Actually, I wouldn't then revert the change. I would change `Bio::EnsEMBL::Hive::Meadow::LOCAL` to whichever meadow is requested, and ask the valley to submit the job

PR #165, opened 2020-06-19
Test init_pipeline outside of the test process and drop the (partial) database if the initialisation fails

muffato:
## Use case

1. `init_pipeline` was the last script tested in the test process itself. A long time ago, I thought it would simplify testing, but I gradually realised it's causing more problems because it is not isolated from the test process. All test functions have since been moved to calling the script with `system` and `init_pipeline` was the last one to work the old way.
One problem is that it loads all the Runnables and registers the pipeline in TheApiary. And since the main script was not tested, some tests were explicitly not using the test function in order to test the script.

2. The other thing I'm addressing here is the ability to expect script failures and to test the error message.

3. If there is say a syntax error in a Runnable, `init_pipeline.pl` rightly detects it:
    ```
    $ init_pipeline.pl Bio::EnsEMBL::Compara::PipeConfig::DumpMultiAlign_conf $(mysql-ens-compara-prod-5-ensadmin details hive) -compara_db $(mysql-ens-compara-prod-9-ensadmin details url muffato_salmonids_epo_101) -division vertebrates -mlss_id 1894 -split_by_chromosome 0 -rel_suffix b
    > Parsing the command-line options.
    > Running the creation commands.
    > Parsing the PipeConfig file and adding objects (this may take a while).
    The runnable module 'Bio::EnsEMBL::Compara::RunnableDB::DumpMultiAlign::MLSSJobFactory' cannot be loaded or compiled:
    Global symbol "$mlss_id" requires explicit package name (did you forget to declare "my $mlss_id"?) at /homes/muffato/workspace/src/rel101/ensembl-compara/modules/Bio/EnsEMBL/Compara/RunnableDB/DumpMultiAlign/MLSSJobFactory.pm line 98.
    Global symbol "$mlss_id" requires explicit package name (did you forget to declare "my $mlss_id"?) at /homes/muffato/workspace/src/rel101/ensembl-compara/modules/Bio/EnsEMBL/Compara/RunnableDB/DumpMultiAlign/MLSSJobFactory.pm line 104.
    Compilation failed in require at (eval 41) line 2.
    ```
    However, this leaves the database in a partial state and I need to add `-hive_force_init 1` once I have fixed the Runnable.

## Description

1/2. I have reimplemented the test `init_pipeline` function to use `_test_ehive_script`, which now has an extra parameter named `$flags` with test options. At this time the only option is `expect_failure` which is either 1 or the regular expression to match the script's standard error against.
3. I made `init_pipeline.pl` drop the database if the initialisation is not complete, which makes it quicker for the user.

## Possible Drawbacks

Perhaps some users manipulate TheApiary in tests after calling `init_pipeline`. They would have to add a line of code to load the pipeline from the database.

## Testing

Yes

Yes

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Scripts/InitPipeline.pm
+++ b/modules/Bio/EnsEMBL/Hive/Scripts/InitPipeline.pm
@@ -59,7 +59,18 @@ sub init_pipeline {
     $hive_dba->dbc->requires_write_access();
 
     print "> Parsing the PipeConfig file and adding objects (this may take a while).\n\n";
-    $pipeconfig_object->add_objects_from_config( $pipeline );
+    eval {
+        $pipeconfig_object->add_objects_from_config( $pipeline );
+    };
+    if ($@) {
+        print "\n> Error: $@\n";
+        print "> Dropping the database due to errors.\n\n";
+        my $drop_cmd = $pipeconfig_object->db_cmd('DROP DATABASE IF EXISTS');
+        if (system($drop_cmd)) {
+            die "Can't drop the database ".$pipeconfig_object->pipeline_url();
+        }
+        exit 1;
+    }
 
     if($tweaks and @$tweaks) {
         print "> Applying tweaks.\n\n";
--- a/modules/Bio/EnsEMBL/Hive/Utils/Test.pm
+++ b/modules/Bio/EnsEMBL/Hive/Utils/Test.pm
@@ -25,6 +25,7 @@ use warnings;
 no warnings qw( redefine );
 
 use Exporter;
+use Capture::Tiny 'tee_stderr';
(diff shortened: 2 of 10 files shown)


ens-bwalts (commented):
This PR has been open so long that peekJob got added after it was submitted. Perhaps update the documentation here to indicate that flags can be passed through to `_test_ehive_script` as well

muffato:
Done. I have rebased the branch and added the doc for `peekJob` and `tweak_pipeline`

muffato:
I have added there a few commits to deal with dropping the database when `init_pipeline` fails

ens-bwalts (approved):

EbiArnie:
Hive is on its way out, this is not getting updated or merged any more.

PR #166, opened 2020-06-19
Cleaner Python code

muffato:
## Use case

As Ensembl is doing more Python there will be more scrutiny on eHive's Python wrapper. `pylint` is a popular Python linter. It complains quite a lot about our code :see_no_evil: as we are not following Python's best practices

## Description

I'm not (yet) keen on addressing every complain, especially as it would change the file history and make it more difficult to trace back the origin of the implementation. However, I think that some changes should still be made in order not to be too far away from the Python world.

I have applied these changes:
- the eHive internal modules are all named in lower case. I haven't changed the main package name `eHive` and the example runnables because that would be a breaking change.
- All internal modules have a dosctring at the beginning
- I broke down the `__all__` line of `__init__.py` to make it easier to maintain (see the next PR #167)
- I have cleaned and ordered the import statements
- I have turned the tests of the Params class into an actual TestCase that is run on Travis
- I have removed `(object)` from the class definition because it's not needed in Python 3
- Slight refactoring of the handling of _tracebacks_ and `validate_parameter_name`

Here are some pylint tests that I have not applied:
- trailing-whitespace
- line-too-long
- bad-whitespace
- trailing-newlines
- unnecessary-pass
- no-else-return
- eval-used

Note that this branch currently includes two commits from #160. I'll remove them once the PR is integrated

## Possible Drawbacks

The functionality is the same.
The per-line history will have changed, but I have tried not to change it too much

## Testing

Yes

Yes

Yes

### DIFF
--- a/scripts/dev/travis_run_tests.sh
+++ b/scripts/dev/travis_run_tests.sh
@@ -32,7 +32,7 @@ else
 fi
 rt=$?
 
-(cd wrappers/python3; python3 -m unittest -v eHive.Process)
+(cd wrappers/python3; python3 -m unittest -v eHive.process)
 rtp=$?
 (cd wrappers/java; mvn "-Dmaven.repo.local=$HOME/deps/maven" test)
 rtj=$?
--- a/scripts/dev/travis_run_tests.sh
+++ b/scripts/dev/travis_run_tests.sh
@@ -32,7 +32,7 @@ else
 fi
 rt=$?
 
-(cd wrappers/python3; python3 -m unittest -v eHive.process)
+(cd wrappers/python3; python3 -m unittest -v eHive.process eHive.params)
 rtp=$?
 (cd wrappers/java; mvn "-Dmaven.repo.local=$HOME/deps/maven" test)
 rtj=$?
--- a/wrappers/python3/README.md
+++ b/wrappers/python3/README.md
@@ -10,6 +10,6 @@ of the same name. The class must inherit from eHive.BaseRunnable (see
 eHive.examples.LongMult.DigitFactory for an example) and implement the
 usual `fetch_input()`, `run()`, and / or `write_output()` methods.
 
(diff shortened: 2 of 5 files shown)


ens-bwalts:
Sounds like this is long overdue, and is a nice set of changes striking a good balance between needed improvements without completely wiping out the history. Certainly it's good to keep this in line with Python coding standards, and there are helpful refactorings here (e.g. moving the exceptions into validate_parameter_name)

Two questions:
1. In params.py, was substitute_param supposed to replace the param_substitute call in internal_get_param? 
2. Is the proposed strategy to merge #167 into this, then re-check that all is OK before merging this into master? Or were you planning to do the merge locally and bring both sets of changes in as one large PR?

muffato:
> 1. In params.py, was substitute_param supposed to replace the param_substitute call in internal_get_param?

No, it wasn't. It's correct as it is. `substitution_in_progress` shouldn't be reset within the circle of `internal_get_param` / `param_substitute` / `subst_all_hashpairs` / `subst_one_hashpair`. But I can see how confusing the names are. `param_substitute` is the equivalent of `Bio::EnsEMBL::Hive::Params::param_substitute` so I would prefer to keep its name. But maybe I should rename `substitute_params` to `substitute_string` because it takes a string as parameter, not a parameter name ?

> 2. Is the proposed strategy to merge #167 into this, then re-check that all is OK before merging this into master? Or were you planning to do the merge locally and bring both sets of changes in as one large PR?

I would suggest this:
1. Merge this one into `master`
2. Change the base branch of #167 to `master`. Hopefully it should detect that these 17 commits are already in and keep on reporting the same diff
3. Merge #167 into `master`

muffato:
- I have renamed `substitute_param` to `substitute_string`
- The Doxygen documentation looks fine

ens-bwalts (approved):

PR #167, opened 2020-06-19
Pure-Python Runnable tester

muffato:
## Use case

In Ensembl/ensembl-py#1 @marcoooo expressed the need to have a pure-Python way of testing Runnables. Part of it was the ability to set up an environment with eHive's Python modules (#157). The other half is being able to run unit-tests without having the Perl side involved.

This PR introduces a way of testing Runnables entirely in Python

:warning: the base branch is itself another PR (#166). I have split both to simplify the review. We can change the base branch for merging later.

## Description

The main eHive module has an extra function named `testRunnable`. Its interface is almost the same as Perl's `standaloneJob` method: it takes the Runnable, the input parameters and the expected list of events raised by the Runnable. The only difference is that it needs a reference to the current `unittest.TestCase`

It works by creating a subclass of the Runnable that overrides the life cycle and all the API methods that can be called, so that it never attempts to read from / send to a Perl side via the JSON stringification.

There are differences, though, and we need to decide whether we allow slight differences between the Perl and Python methods. Generally, I have tried to make the interface more Pythonic

1. Events are indicated as instances of lightweight classes available through the eHive module, rather than an array-ref whose first element is the type of the event. This allows checking that the arguments are correct.
2. I have added an event for CompleteEarly. I could add it in Perl too, but it would break all the existing Perl Runnable tests (in all repos, not only ensembl-hive) that currently don't have it
3. Failure is checked through an event too, and Python allows checking that the actual exception type is correct. In Perl, this is done through the `expect_failure` boolean flag
4. I have added an option to test the Runnable in retry mode (in case it behaves differently when `retry_count` is not 0)
5. I have added a mechanism to test the final value of `autoflow`, `lethal_for_worker` and `transient_error`. These can't (yet ?) be tested in Perl.

## Possible Drawbacks

None, it's a new feature

## Testing

Yes, the example Runnable `TestRunnable` has some unit-tests

Yes

Yes


muffato:
This branch has been rebased and I have added a commit to use `doxypypy` on ReadTheDocs instead of `doxypy`. It handles Python better than `doxypy` does. See https://github.com/Feneric/doxypypy

muffato:
See https://ensembl-hive.readthedocs.io/en/tmp-readthedocs_test/python3/namespaceeHive_1_1tests.html#a32b710fe9be9d81300954d7b47da8bf6 : it can parse the list of arguments and format them better

muffato:
Rebased to clear the conflict warnings

PR #168, opened 2020-07-03
bugfix: the version of sqlite3 for Ubuntu 18.04 is 1.56, not 1.58

muffato:
## Use case

I've found out that the readthedocs are failing (on both 2.5 and master) because this sqlite .deb file can't be found

## Description

According to https://packages.ubuntu.com/bionic/libdbd-sqlite3-perl, the current version on Ubuntu 18.04 (the one used on readthedocs) is 1.56, not 1.58

## Possible Drawbacks

None

## Testing

ReadTheDocs built successfully: https://readthedocs.org/projects/ensembl-hive/builds/11385254/


ens-bwalts:
Changes merged with #169

PR #171, opened 2020-08-21
One hash fix

MatBarba:
## Use case

When a parameter contains a string with a single literal "#" symbol in it, the python3 wrapper tries to find a second "#" to do a substitution, and then dies because it cannot find one. This can be the case for e.g. a strain name that contains a "#".

## Description

This simple workaround allows a single "#" symbol to be used within a string as a parameter value when using python runnables.

It doesn't solve the issue in case several literal "#" symbols are in the same string, and it still doesn't allow having substitutions + literal "#". A more general escape mechanism for this special character would be needed, or the behaviour should be documented.

## Possible Drawbacks

The runnables do not fail any more if a placeholder to be substituted only has one "#" instead of two.

## Testing

I've added a substitution test in the TestRunnable.py, with a string parameter with one hash.

I have also run several pipelines with this fix and it solved this rare issue without impacting the normal substitutions.

### DIFF
--- a/wrappers/python3/eHive/Params.py
+++ b/wrappers/python3/eHive/Params.py
@@ -160,7 +160,7 @@ class ParamContainer(object):
         """
         self.debug_print("subst_all_hashpairs", structure)
         
-        # Allow a single literal hash?
+        # Allow a single literal hash
         if structure.count("#") == 1:
             return structure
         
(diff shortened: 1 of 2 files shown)


ens-bwalts (approved):
This works as a stopgap solution to get the Python wrappers working more like the Perl wrappers.

muffato:
Note: the fix should be backported to `version/2.4`

ens-bwalts:
Superceded by #172 - after merging into version/2.4 a cascade merge will bring this up through version/2.5 and master

PR #174, opened 2020-11-09
Bugfix LSF detection

muffato:
## Use case

We are testing some pipelines in a Slurm farm that provides `lsid`, and eHive activates the LSF meadow tries to submit jobs with `bsub` etc, which doesn't exist, and fails.
A workaround is to run beekeeper with `--meadow_type SLURM`, but I believe the LSF meadow should be fixed.

## Description

Slurm's `lsid` does print it is a Slurm, it doesn't pretend to be an LSF. The fix is change the parsing of `lsid` to look for the LSF keyword, and openlava too since they are mostly the same.

## Possible Drawbacks

None envisaged.

## Testing

Yes, `lsf.t` now tests the meadow detection against a collection of `lsid` outputs

They do

All good (but remember that we don't have Travis any more !)

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Meadow/LSF.pm
+++ b/modules/Bio/EnsEMBL/Hive/Meadow/LSF.pm
@@ -54,12 +54,16 @@ our $VERSION = '3.1';       # Semantic version of the Meadow interface:
 =cut
 
 sub name {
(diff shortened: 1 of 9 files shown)


ens-bwalts (approved):
Discussed this with @carlacummins - this is a much more robust way to check that we are on the correct meadow

PR #175, opened 2020-11-10
Code maintenance - RunnableDB::DatabaseDumper

muffato:
## Use case

A few changes to improve maintainability of `RunnableDB::DatabaseDumper`

## Description

1. Instead of setting the command to `true` to breeze through `run` and `write_output`, use `$self->complete_early()` (which was introduced after the creation of this Runnable
2. Don't add the "signature" if the Runnable is run in standalone mode,

## Possible Drawbacks

None envisaged

## Testing

There are no tests for this Runnable


ens-bwalts (approved):
Yeah, replacing the mysqldump command with true is a bit obscure, this is much more intuitive.

PR #177, opened 2020-11-18
Drop error on null hash key in accu

mira13:
## Use case

When hash key is set to null in dataflow
## Description

When hashkey is set to null, next job fails to read it, but it should be that current job fails to read it

## Possible Drawbacks

Not found

## Testing

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/Accumulator.pm
+++ b/modules/Bio/EnsEMBL/Hive/Accumulator.pm
@@ -1,4 +1,4 @@
-=pod 
+=pod
 
 =head1 NAME
 
@@ -115,6 +115,8 @@ sub dataflow {
             my $key_signature = $accu_address;
             $key_signature=~s/(\w+)/$emitting_job->_param_possibly_overridden($1,$output_id)/eg;
 
+            $self->checkEmptyKeys($key_signature, $accu_address);
+
             push @rows, {
                 'sending_job_id'            => $sending_job_id,
                 'receiving_semaphore_id'    => $receiving_semaphore_id,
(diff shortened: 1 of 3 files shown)


ens-bwalts (commented):
I'm pretty sure that this is all that is needed (see if we end up with an empty {} of a key signature, when that wasn't what was intended by comparing to the original accu address)

This needs a test - there aren't many accu tests in eHIve - other than the integration tests in longmult.t, gcpct.t, and kmer.t) so this could be a good time to bring them in - possibly in a new test case under 10.pipeconfig.

muffato:
The problem happens on hashes _anywhere_ in the address, not just when the address is entirely an hash. For instance, take a look at the k-mer example pipeline [KmerPipelineHoA_conf.pm#L262](https://github.com/Ensembl/ensembl-hive/blob/version/2.5/modules/Bio/EnsEMBL/Hive/Examples/Kmer/PipeConfig/KmerPipelineHoA_conf.pm#L262).
The address is `{kmer}[]`, and if `kmer` is missing from the dataflow in [CountKmers.pm#L133](https://github.com/Ensembl/ensembl-hive/blob/version/2.5/modules/Bio/EnsEMBL/Hive/Examples/Kmer/RunnableDB/CountKmers.pm#L133) the `accu` table gets data like this
```
+----------------+------------------------+-------------+---------------+-------+
| sending_job_id | receiving_semaphore_id | struct_name | key_signature | value |
+----------------+------------------------+-------------+---------------+-------+
|              4 |                      1 | all_counts  | {}[]          | 8     |
|              4 |                      1 | all_counts  | {}[]          | 8     |
|              4 |                      1 | all_counts  | {}[]          | 8     |
|              4 |                      1 | all_counts  | {}[]          | 8     |
|              4 |                      1 | all_counts  | {}[]          | 8     |
+----------------+------------------------+-------------+---------------+-------+
```
which breaks later on when running the funnel job:
```Worker 4 [ Role 4 , compile_counts(5), Job 2 ] WORKER_ERROR : Can't use string ("41") as an ARRAY ref while "strict refs" in use at /home/matthieu/workspace/src/hive/master/modules/Bio/EnsEMBL/Hive/Examples/Kmer/RunnableDB/CompileCountsHoA.pm line 98.```
as the outer datastructure is now a multiset (hash: key → integer) instead of a hash key → array

File: modules/Bio/EnsEMBL/Hive/Accumulator.pm [hunk 1]
```diff
@@ -131,6 +133,60 @@ sub dataflow {
...
+    }
+  }
+
+  my @squareBrackets = $self->findEmptyBrackets($key_signature, '[', ']');
+  my @squareBracketsAddress = $self->findEmptyBrackets($accu_address, '[', ']');
+  %addressBrackets = map { $_ => 1 } @squareBracketsAddress;
+
+  foreach my $position ( @squareBrackets ) {
+    if (!exists($addressBrackets{$position})) {
+        die "Null hash key in accumulator";
```
ens-bwalts, line 156: This one (for `[]`) should read Null array key. Since you have the positions of the array or hash elements that didn't get substituted, it would be kind to the user to include the original accu_address and the positions that caused the problem in the output message.

Otherwise, it seems like this works. I suspect that it would also be possible to use pattern matching to check this, but your approach has an advantage in readability and maintainability.

ens-bwalts (changes requested):

mira13:
Ready for review

modules/Bio/EnsEMBL/Hive/Accumulator.pm
```diff
@@ -115,6 +115,8 @@ sub dataflow {
             my $key_signature = $accu_address;
             $key_signature=~s/(\w+)/$emitting_job->_param_possibly_overridden($1,$output_id)/eg;
```
muffato, line 116: I was thinking about an alternative solution. First here:
```suggestion
            $key_signature=~s/(\w+)/$self->_check_empty_key($1, $emitting_job->_param_possibly_overridden($1,$output_id))/eg;
```
And something like this further down in the module:
```perl
sub _check_empty_key {
    my ($key, $value) = @_;
    if ((not defined $value) || ($value eq '')) {
        die "No value provided for $key. It is needed for the accu dataflow.";
    }
}
```

The function call is only activated when `\w+` is found, which should be for all hash keys/array indexes.
Something to check though: is the `die` going to break the `s///` substitution and "bubble" up ?

mira13: Regexp is possible, but current solution it's more maintanable and readable, and we give more information for user in error - which exacly brackets brakes a job.

modules/Bio/EnsEMBL/Hive/Accumulator.pm
mira13: @ens-bwalts all done, thanks for review

muffato:
The other (minor) comment I have is that the eHive code is snake_case, not CamelCase

mira13:
@muffato thanks, fixed

mira13:
@ens-bwalts all changes are complete

modules/Bio/EnsEMBL/Hive/Accumulator.pm
```diff
@@ -131,6 +133,60 @@ sub dataflow {
...
+  %addressBrackets = map { $_ => 1 } @$squareBracketsAddress;
+
+  foreach my $position ( @$squareBrackets ) {
+    if (!exists($addressBrackets{$position})) {
+        die "Null array key in accumulator, empty square brackets number " . $position;
+    }
+  }
+}
+
+sub findEmptyBrackets {
```
EbiArnie, line 161: May I suggest something like this:
`sub findEmptyBrackets {
    my ( $self, $string, $open, $close ) = @_;
    my $count = 0;
    my $result = [];

    # look for opening bracket
    while ($string =~ /$open/g) {
        # count how many opening brackets we have
        $count++;
        if ($string =~ /\G(?=$close)/) {
            # store number of bracket if we find an empty pair (like {})
            push (@$result, $count);
        }
    }
    return $result;
}
`

I find the current version a little hard to read.
If you don't like the lookahead assertion (the ?= in the regex), this should be equivalent:
`if (substr($string, pos($string), 1) eq '}');`

sgiorgetti:
Changes merged in with PR #205

PR #178, opened 2020-12-01
[generate_timeline] Define custom categories to group analyses / resource-classes

muffato:
## Use case

James asked us to extract farm usage statistics for the migration to Codon, for some memory classes. I have used `generate_timeline.pl` to get the data for Compara, but did a few changes to allow defining the custom classes he wanted to have (which in our cases consisted of grouping some of our resource classes together).

## Description

`generate_timeline.pl` accepts a new parameter: a script that defines a function named `get_key_name` that, given an Analysis or a ResourceClass object, returns a category name for this object. Objects that have the same category name are grouped together and displayed with the name given. I have added the function that I used for Compara as an example.

I also took the opportunity to rename several variables for clarity.

## Possible Drawbacks

None

## Testing

There are no unit test for `generate_timeline.pl` and I didn't want to embark on writing one

### DIFF
--- a/scripts/dev/generate_timeline_example_key_transform_file.pl
+++ b/scripts/dev/generate_timeline_example_key_transform_file.pl
@@ -1,0 +1,69 @@
+
+use strict;
+use warnings;
+
+# Transform a Compara resource class into a category name according to the
+# memory requirement.
+
+sub get_key_name {
+    my $resource_class = shift;
+
+    my $display_name = $resource_class->display_name;
+    my $memory_req = $display_name;
+
+    # Convert special names to the standard nomenclature
+    $memory_req = '250Mb_job' if $display_name eq 'default';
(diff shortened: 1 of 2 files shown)


ens-bwalts (approved):
The risky bit with the refactoring of generate_timeline.pl was renaming `%key_name` to `%key_name_mapping`, then introducing a new `$key_name` scalar, but it looks like all of the changes were made here. This is a sensible refactoring, combined with a feature add, and includes a nice example of the new key_transform_file

TamaraNaboulsi (approved):
The 'main' function of generate_timeline.pl now accepts a new parameters, key_transform_file, which is the name/path of a script which should contain a 'get_key_name' function. The variable %key_name was also changed to %key_name_mapping and the code was updated to use the 'get_key_name' function to populate this hash. An example script was also added to see how 'get_key_name' can be defined.

Changes look good and clear to me.

PR #180, opened 2020-12-01
Cleaner python wrapper

muffato:
## Use case

More changes I should have put into #166 . They make `wrappers/python3/wrapper` more compliant with `pylint` (although still not 100%) and more Pythonic.

## Description

The biggest change is to use a `namedtuple` to hold the details about each mode of the wrapper instead of a regular tuple. This means that the data can be accessed through named properties `function` and `args` rather than indices `[0]` and `[1]`.

## Possible Drawbacks

None

## Testing

The wrapper is covered by several tests

### DIFF
--- a/wrappers/python3/wrapper
+++ b/wrappers/python3/wrapper
@@ -40,7 +40,6 @@ def do_run():
 
 def do_build():
     print("Nothing to do")
-    return
 
 ## And here we select the mode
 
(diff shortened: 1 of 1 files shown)


ens-bwalts (approved):
namedtuples does indeed make this more readable. Syntax correct.

TamaraNaboulsi (approved):
Using namedtuple for modes wrapper to create tuples that have access to 'args' and 'function' properties. It makes the code more clear by not using [0] and [1] for these properties and makes the code compliant with plyint. Changes look good.

PR #181, opened 2020-12-01
Misc. minor dev

muffato:
## Use case

A collection of minor changes I have locally: bugfixes and new features

## Description

1. bugfix: avoid Perl's warning about undefined values.
2. New SQL view similar to `progress`, which shows which analysis created each job. The view is not enabled yet because I didn't want to create a new schema version just for that right now. If you want to keep the view, a patch can be added now, or later, bundled with another schema change so that users of the `master` branch don't see too many changes.
3. Moved the housekeeping .t tests into their own directory
4. Fixed the executable bit of two .t

## Possible Drawbacks

None

## Testing


EbiArnie:
Hive is on its way out, this is not getting updated or merged any more.

PR #184, opened 2020-12-15
Various bug fixes in tweak_pipeline

muffato:
## Description

Equivalent of #183 but for the `master` branch. This time I have added a test in `tweak_pipeline_output.t`. Remember that the latter only checks that the output JSON conforms to the spec, not that each individual tweak is doing the correct thing (for that you need to run the test in verbose mode and read its output)

## Possible Drawbacks

None

## Testing

Yes

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/HivePipeline.pm
+++ b/modules/Bio/EnsEMBL/Hive/HivePipeline.pm
@@ -854,6 +854,11 @@ sub apply_tweaks {
                            $need_write = 1;
                         }
                     }
+
+                } elsif( $attrib_name eq 'dbID' ) {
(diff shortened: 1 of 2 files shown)


ens-bwalts (approved):
This PR fixes a few longstanding annoyances within the tweak system, mostly regarding resource classes. New resource classes must be explicitly created before being assigned to an analysis. Resource classes can't be removed from an analysis (these are the changes in the low-800 lines of HivePipeline.pm). It also creates resource classes when setting a resource description. 

There is also a bugfix prohibiting the deletion of an analysis' dbID

ens-bwalts:
#190 (functionally similar PR with some changes to prevent merge errors) has an approving review, this has an approving review. Will go ahead and merge these PRs in the interest of expediency.

PR #185, opened 2020-12-15
Added a method to create arbitrary dependency graphs between jobs (using semaphores)

muffato:
## Use case

This is an ability I needed for a project where I had a lot of data to process and condense into 1 output file. I used this function to create jobs that will process (merge) the input files 5 at a time, then process their outputs 5 at a time, etc, recursively, until there is a single output file. I could probably have done that with regular dataflows, but that would have involved a lot of Dummy / Factory jobs to control the jobs and create the dependencies.

## Description

I have added a method to create arbitrary dependency graphs between jobs (using semaphores). It needs a hash that describes the dependencies and can be simpler to use than the PipeConfig language.

It allows escaping from semaphores and can be seen as a solution to ENSCORESW-3319.
It could also be extended to implement other features people have asked about semaphores, such as a job being involved in multiple semaphores.

## Possible Drawbacks

There is a very short window of time where the funnel would be in READY state, because it is created with no fan jobs blocking it (the fan jobs being added later). That wasn't an issue because my beekeeper was not looping at the time but who knows, a worker could claim jobs at the same time. I think it can be fixed by adding some options in `store_a_semaphored_group_of_jobs` or, copying it into `store_nested_semaphores` but neither solution seems to be as clean as I'd hoped.

## Testing

Yes, there is a new unit test

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Hive/DBSQL/AnalysisJobAdaptor.pm
@@ -355,6 +355,80 @@ sub store_a_semaphored_group_of_jobs {
 }
 
 
+=head2 store_nested_semaphores
+
+  Arg [1]    : Hashref $semaphore_hash
+  Arg [2]    : (optional) Bio::EnsEMBL::Hive::AnalysisJob $emitting_job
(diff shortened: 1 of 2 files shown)


sgiorgetti:
Converting to draft PR for now and keep this for future reference.

PR #188, opened 2021-06-08
Checks for ENSEMBL_ROOT_DIR and ENSEMBL_CSV_ROOT_DIR

TamaraNaboulsi:
## Description

EnsemblGeneric_conf checks for installed Ensembl modules in $ENSEMBL_CVS_ROOT_DIR, but these modules will be stored in $ENSEMBL_ROOT_DIR on Codon. This change ensures that we check both directories, with the following priorities in place:

- 'ensembl_root_dir' passed on the command line
- 'ensembl_cvs_root_dir' passed on the command line
- $ENSEMBL_ROOT_DIR environment variable
- $ENSEMBL_CVS_ROOT_DIR environment variable

JIRA ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-3735

## Possible Drawbacks

This fix gives values passed through the command line priority over environment variables, but it was the reverse before. Thus, this fix changes the behavior of the code so users should be aware of that. However, the reverse of this was deemed a bug and users have asked for this fix so all should be good.

## Testing

No (I don't think there is any test that checks this param)

### DIFF
--- a/modules/Bio/EnsEMBL/Hive/PipeConfig/EnsemblGeneric_conf.pm
+++ b/modules/Bio/EnsEMBL/Hive/PipeConfig/EnsemblGeneric_conf.pm
@@ -61,7 +61,7 @@ sub default_options {
             #   [bash]      export -n ENSEMBL_CVS_ROOT_DIR  # will stop exporting, but the value in current shell stays as it was
             #   [tcsh]      unsetenv ENSEMBL_CVS_ROOT_DIR   # will destroy the variable even in current shell, and stop exporting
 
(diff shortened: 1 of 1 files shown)


ens-bwalts (approved):
Priority is now sensible, and matches the requests from Slack. Options set via the command line (e.g. `$self->o()`) now override environment variables.

nwillhoft (approved):
This PR updates the priority order so that values from the command line are checked before values from environmental variables.  This update has been requested by ensembl-hive users as a bug fix and all looks good.

PR #189, opened 2021-06-08
EnsemblGeneric_conf checks for ENSEMBL_ROOT_DIR and ENSEMBL_CSV_ROOT_DIR

TamaraNaboulsi:
## Use case

Codon guidelines call for the Ensembl installation path to be stored in $ENSEMBL_ROOT_DIR. EnsemblGeneric_conf needs to now check this before checking $ENSEMBL_CVS_ROOT_DIR in case it is being run on Codon.

## Description

This change ensures that EnsemblGeneric_conf checks both $ENSEMBL_ROOT_DIR and $ENSEMBL_CVS_ROOT_DIR, giving priority to $ENSEMBL_ROOT_DIR and uses whichever one it finds first.

JIRA ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-3735

## Possible Drawbacks

None

## Testing

No (I don't think there is any test that checks this)


ens-bwalts (changes requested):
This looks correct, but could you please make the base branch version/2.5.

TamaraNaboulsi:
> 
> 
> This looks correct, but could you please make the base branch version/2.5.

I did it against version/2.5 first but the checks were failing and I thought that was the reason. Will reopen that PR then. Thanks Brandon!

PR #195, opened 2021-08-23
Making bacct options configurable according to cluster

TamaraNaboulsi:
## Requirements

- Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion;
- Review the [development guidelines](https://ensembl-hive.readthedocs.io/en/master/dev/development_guidelines.html#ehive-development-in-a-nutshell) for eHive; remember in particular:
    - Do not modify code without testing for regression.
    - Provide simple unit tests to test the changes.
    - If you change the database schema, please follow the instructions [for schema changes in the developer guidelines](https://ensembl-hive.readthedocs.io/en/master/dev/development_guidelines.html#schema-changes).
    - If you change the schema, meadow, or guest language interfaces, please follow the scheme [for internal versioning in the developer guidelines](https://ensembl-hive.readthedocs.io/en/master/dev/development_guidelines.html#internal-versioning).
    - The PR must not fail unit testing.

## Use case

The bacct command takes in an extra option (-f -) on Codon. Adding this option doesn't currently hurt anything on Noah but future additions might. Thus, it is better to make these extra options configurable per cluster in case some more tuning is needed for the bacct command.

JIRA ticket: [ENSCORESW-3806](https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-3806)

## Description

Added a new config option "BacctExtraOptions" which has the value if "-f -" for codon. This config value is read the placed in the bacct command in addition to other options.

## Possible Drawbacks

None

## Testing

The existing test 04.meadow/lsf.t was modified to read the bacct extra options and it is succeeding with this change.

### DIFF
--- a/hive_config.json
+++ b/hive_config.json
@@ -36,6 +36,9 @@
             "EBI" : {
                 "SubmissionOptions"  : "",
                 "TotalRunningWorkersMax"    : 10000
+            },
+            "codon" : {
+                "BacctExtraOptions"  : "-f -"
             }
         },
         "SGE" : {
--- a/modules/Bio/EnsEMBL/Hive/Meadow/LSF.pm
+++ b/modules/Bio/EnsEMBL/Hive/Meadow/LSF.pm
@@ -322,7 +322,8 @@ sub get_report_entries_for_process_ids {
 
     unless ($self->config_get('AccountingDisabled')) {
         while (my $pid_batch = join(' ', map { "'$_'" } splice(@_, 0, 20))) {  # can't fit too many pids on one shell cmdline
-            my $cmd = "bacct -f - -l $pid_batch";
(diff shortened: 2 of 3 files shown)


ens-bwalts (approved):
Changes look good - adding a BacctExtraOptions to hive_config.json and fetching them from the config file, with a no-option default. 

I tested this using the `Bio::EnsEMBL::Hive::Examples::FailureTest::PipeConfig::MemlimitTest_conf` pipeline on both noah and codon. This pipeline causes jobs to fail due to running out of memory. These failures are detected by the Meadow looking at the output from running the `bacct` command. If working correctly, after the pipeline runs you should be able to see workers with a cause_of_death of "MEMLIMIT" in the worker table of the hive database. I saw this for pipelines run on both noah and codon.

PR #205, opened 2022-04-28
Revisited 'Drop error on null hash key in accu , PR #177'

EbiArnie:
## Use case

When hash key is set to null in dataflow

## Description

When hashkey is set to null, next job fails to read it, but it should be that current job fails to read it

## Possible Drawbacks

Not found

## Testing

Yes


sgiorgetti (approved):
Looks good to me.
Please @EbiArnie - go ahead and prepare to merge these also into `main`. Thx

PR #210, opened 2023-05-03
Remove redundant message

MatBarba:
Remove a redundant message in the Slurm meadow that is already printed by beekeeper.


EbiArnie (approved):
OK

PR #211, opened 2023-10-24
Mbarba/fix capture error

MatBarba:
## Use case

The die within the capture after the system call uses both $! and $?. If the command fails,
this might raise some errors like:

`Could not submit job(s): Inappropriate ioctl for device, 256 at ...`

The actual problem is not displayed.

## Description

Remove the die() within the capture. If the command fails, the stderr and stdout are properly captured afterwards and displayed.

## Possible Drawbacks

Unlikely.

## Testing

Run the test suite, no difference with previous commit.


EbiArnie (approved):
OK

PR #212, opened 2023-11-17
Add pyproject.toml File

JAlvarezJarreta:
## Use case

Our CI/CD system has flagged a future problem with the current Python configuration system:
```
DEPRECATION: ensembl-hive is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
```
The current `pip` version installed is 23.0.1 so this issue will soon be affecting many CI/CD systems.

## Description

I have moved the current `setup.py` configuration into `pyproject.toml` to avoid the issue mentioned above. I have noticed the `__version__` said 5.0 so I have changed it to 2.6 to match eHive's version.

## Possible Drawbacks

None.

## Testing

I was installed the library using this new configuration and the result is the same as in the previous installation.


muffato:
Hi @JAlvarezJarreta !

That version number in `wrappers/python3/eHive/process.py` (5.0) has a meaning of its own. In eHive, "guest languages", like Python, are plugins and eHive uses the version number to make sure the plugin it finds is compatible.
The versioning follows semantic versioning. The version 5 was introduced in this commit https://github.com/Ensembl/ensembl-hive/commit/4e6597cdaf2e8b14ae00fb1bc8fdb314c59e92ec , which introduced a breaking change and required the two guest languages Java and Python to have their code updated + version bumped accordingly.
It's the same principle as the "meadow" interface versioning to connect LSF, SLURM, etc, to eHive - which happens to also be version 5 at the moment, but that's just a coincidence. All those versions can be displayed by beekeeper:

```
$ beekeeper.pl --versions
CodeVersion     2.6
CompatibleHiveDatabaseSchemaVersion     96
MeadowInterfaceVersion  5
Meadow::LOCAL   5.0     available
Meadow::LSF     5.2     available
GuestLanguageInterfaceVersion   5
GuestLanguage[java]     N/A     incompatible
GuestLanguage[python3]  5.0     available
```
(I haven't compiled the Java wrapper, that's why it's marked as N/A).

I haven't checked but I suspect that's why the CI fails.

JAlvarezJarreta:
> Hi @JAlvarezJarreta !
> 
> That version number in `wrappers/python3/eHive/process.py` (5.0) has a meaning of its own. In eHive, "guest languages", like Python, are plugins and eHive uses the version number to make sure the plugin it finds is compatible. The versioning follows semantic versioning. The version 5 was introduced in this commit [4e6597c](https://github.com/Ensembl/ensembl-hive/commit/4e6597cdaf2e8b14ae00fb1bc8fdb314c59e92ec) , which introduced a breaking change and required the two guest languages Java and Python to have their code updated + version bumped accordingly. It's the same principle as the "meadow" interface versioning to connect LSF, SLURM, etc, to eHive - which happens to also be version 5 at the moment, but that's just a coincidence. All those versions can be displayed by beekeeper:
> 
> ```
> $ beekeeper.pl --versions
> CodeVersion     2.6
> CompatibleHiveDatabaseSchemaVersion     96
> MeadowInterfaceVersion  5
> Meadow::LOCAL   5.0     available
> Meadow::LSF     5.2     available
> GuestLanguageInterfaceVersion   5
> GuestLanguage[java]     N/A     incompatible
> GuestLanguage[python3]  5.0     available
> ```
> 
> (I haven't compiled the Java wrapper, that's why it's marked as N/A).
> 
> I haven't checked but I suspect that's why the CI fails.

Ah, I didn't know. Thank you for the information! Will revert back the version change then. :+1:

sgiorgetti (approved):
OK

PR #215, opened 2024-01-12
Restrain using DBD::mysql version 5.x

sgiorgetti:
## Description

DBD::mysql have released new version (5.001) on Oct 4, 2023. This version removes support for MySQL 5 -
https://metacpan.org/dist/DBD-mysql/changes

Hence we need to use lower version than that.

## Use case

Using DBAdaptors to connect to Ensembl databases which use MySQL version 5

## Possible Drawbacks

Should be changed back if anytime in the future Ensembl database servers upgraded to MySQL 8x.

## Testing

Unit test should pass on this PR

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -1,5 +1,5 @@
 requires 'DBI';
-requires 'DBD::mysql';
+requires 'DBD::mysql', '<= 4.050'; # newer versions do not support MySQL 5
 requires 'DBD::SQLite';
 requires 'DBD::Pg';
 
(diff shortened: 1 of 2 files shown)


cpanfile
```diff
@@ -1,5 +1,5 @@
 requires 'DBI';
-requires 'DBD::mysql';
+requires 'DBD::mysql', '<= 4.050'; # newer versions do not support MySQL 5
```
EbiArnie, line 2: We should use 4.051 here.
With that OK from me.

EbiArnie (changes requested):

cpanfile
sgiorgetti: Changed to comply and accommodate further possible minor updates.

EbiArnie (approved):
OK

PR #217, opened 2024-01-17
Fix RTD

sgiorgetti:
## Use case

RTD build is failing because `os.image` directive has been deprecated and the mandatory `build.os` is not found.
[LINK](https://blog.readthedocs.com/use-build-os-config/)

## Possible Drawbacks

None

## Testing

No changes were necessary to the eHive code.
RTD is not covered by the test suite, but local creation of docs worked fine, as well as on RTD.


EbiArnie (approved):
It's nice having docs again. OK!

PR #219, opened 2024-07-11
Add eHive REST Client

JAlvarezJarreta:
## Description

Moving `HiveRESTClient` code from ensembl-py to its rightful placement, including its unit test. I have taken the opportunity to update `pyproject.toml` and move the minimum supported Python version to 3.10.

## Possible Drawbacks

None (to my knowledge).

## Testing

Unit test provided passes successfully.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,12 @@ dependencies = [
     "requests >= 2.22.0",
 ]
 
+[project.optional-dependencies]
+cicd = [
+    "pytest",
+    "requests-mock >= 1.8.0",
+]
+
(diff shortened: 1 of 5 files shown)


EbiArnie:
This is nice, but does anyone use this REST client?

JAlvarezJarreta:
This was implemented by Production (Marc), so I must confess I have no clue. Wanted to preserve it just in case, but happy with whatever you decide :relaxed:

EbiArnie:
I could not find any users of this code.
I will drop this for now. We can always resurrect it if needed.

PR #222, opened 2024-09-24
Update readthedocs build

EbiArnie:
Somehow, the build file for RTD was not picked up when merging the branches.
This should fix the build.


sgiorgetti (approved):
same as per v2.6 - LGTM

PR #225, opened 2024-09-25
Check for undefined config hash

EbiArnie:
Check for undefined config hash before setting config.
Some teams instantiate a Valley without config to query which Meadows are available.


sgiorgetti (approved):
OK

PR #226, opened 2024-09-25
Add recommonmark for rtd

EbiArnie:
- readthedocs no longer installs some dependencies. Add them explicitly.

They also mention mock, pillow and commonmark, but I believe we do not need those.


sgiorgetti (approved):
OK

PR #232, opened 2024-10-16
Updated docs because of meadow deprecations

sgiorgetti:
## Use case

We have deprecated eHive support for some job scheduler, and consequently eHive meadows.
Docs need updating to reflect these changes.

## Possible Drawbacks

None

## Testing

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,14 @@ eHive is a system for running computation pipelines on distributed computing res
 
 The name comes from the way pipelines are processed by a swarm of autonomous agents.
 
+> [!IMPORTANT]  
+> As per eHive version 2.7.0, we changed the preferred/default eHive meadow from *LSF* to *SLURM*.
+>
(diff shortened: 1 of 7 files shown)


sgiorgetti:
@TamaraNaboulsi  - would you mind to have a look at this?
The majority of examples in the entire doc set is based on LSF for historical reasons.
TBH, I applied minimum changes and unsure if adequate - especially on `docs/running_pipelines/management.rst`.
Does it make sense? Pls, comment and criticise at will. Thanks

EbiArnie (approved):
Definitely and improvement over the status quo.
OK from my side.

TamaraNaboulsi (approved):
Very much needed docs update now that SLURM is the default rather than LSF. Changes look good and are very descriptive.
Concerning "management.rst", it does still contain multiple mentions of LSF. However, these mentions are used in examples of hive operations (like beekeeper output) and thus the actual information there doesn't change whether it is LSF or SLURM. It would be nice to change these later on just so the operations of the users look like the examples here, but it's not super important for now.

PR #234, opened 2025-08-27
Adapt to changes in sinfo output

EbiArnie:
- sinfo does not print the slurmd version for each partition, instead giving N/A. We now run 'sinfo -V' separately to get the version.


sgiorgetti (approved):
LGTM

